PR #7, opened 2022-07-18
expose libusb_claim_interface and release_interface

rburanyi:
The current libhoth implementation does not support multiple
applications talking to the same device at the same time, this can be
fixed by claiming/releasing the USB device periodically.

Signed-off-by: Robi Buranyi <rburanyi@google.com>

### DIFF
--- a/examples/htool_console.c
+++ b/examples/htool_console.c
@@ -61,18 +61,18 @@ static int get_channel_status(struct libhoth_usb_device *dev,
   }
 
   *offset = resp.write_offset;
-  return 0;
+  return LIBHOTH_OK;
 }
 
 static int read_console(struct libhoth_usb_device *dev,
                         const struct htool_console_opts *opts,
-                        uint32_t *offset) {
+                        uint32_t *offset, size_t *response_size) {
   struct ec_channel_read_request req = {
       .channel_id = opts->channel_id,
       .offset = *offset,
       .size =
           HOTH_FIFO_MAX_REQUEST_SIZE - sizeof(struct ec_channel_read_response),
-      .timeout_us = 10000,
+      .timeout_us = 0,
   };
 
   struct {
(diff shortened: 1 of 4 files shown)


libhoth_usb.c
```diff
@@ -98,10 +98,6 @@ int libhoth_usb_open(const struct libhoth_usb_device_init_options* options,
   if (status != LIBUSB_SUCCESS) {
     goto err_out;
   }
-  status = libusb_claim_interface(dev->handle, info.interface_id);
```
korran, line 101: This will break what the hothd folks are currently doing. Here's how I was originally thinking this should work so that it doesn't change the API semantics:

- libusb_claim_interface() increments a reference count. If the previous reference count was zero, it asks the kernel to claim the interface

- libusb_release_interface() decrements a reference count, and when zero asks the kernel to release the interface.

- libhoth_usb_send_request() calls libusb_claim_interface() (plus libusb_release_interface() on error), and libhoth_usb_receive_response() calls libusb_release_interface after it has received the response.

examples/htool_console.c
```diff
@@ -222,10 +231,14 @@ int htool_console_run(struct libhoth_usb_device *dev,
...
   bool quit = false;
   while (!quit) {
-    status = read_console(dev, opts, &offset);
+    size_t response_size;
+    status = read_console(dev, opts, &offset, &response_size);
     if (status != LIBHOTH_OK) {
       break;
     }
+    if (response_size == 0) {
+      usleep(10000);
```
korran, line 240: Can I propose something a bit different?

What do you think of having a new option to the console, --log, which does the following:

- stops listening to stdin
- polls once a second instead of once every 10ms
- pastes a timestamp after every newline character

I can implement this if you like...

korran: The other thing you might want it to do is to retry when dauntless restarts and re-enumerates over USB

rburanyi: Thanks for the review. Yes, "--log" looks good and would make a lot of sense if it did not read from stdin. Thanks in advance if you can do that.
I do not need a timestamp after every newline character. I did implement that earlier (I still have the patch handy) but turns out I can also do that with simply piping through `ts +%s`.

rburanyi: "retry when the RoT restarts" ... I can currently retry in the script so when `htool console` terminates I can just respawn it, because the USB address is anyway changing.

cjevans-google:
Obsoleted by https://github.com/google/libhoth/pull/36

PR #10, opened 2022-07-29
Add support for sending RS-232 break sequence.

korran:
See https://en.wikipedia.org/wiki/Universal_asynchronous_receiver-transmitter#Break_condition


rburanyi (approved):

cjevans-google:
Closed as obsoleted: We were far too slow to review this change.  Possibly it's still desireable, but we need to revisit the feature.

PR #12, opened 2022-09-07
build: Fixup for suitable usage as a meson subproject

wak-google:
This cleans up the build structure so that we can use the `hoth_usb_dep` in meson subprojects to leverage libhoth as a dependency. This also disables building htool by default when being used a subproject library.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,12 +1,49 @@
-project('libhoth', 'c', 'cpp', license: 'Apache-2.0', version: '0.0.0')
+project(
+  'libhoth',
+  'c',
+  license: 'Apache-2.0',
+  version: '0.0.0',
+  default_options: [
+    'examples=' + (meson.is_subproject() ? 'false' : 'true'),
+  ])
 
-libhoth_usb_srcs = ['libhoth_usb.c', 'libhoth_usb_fifo.c', 'libhoth_usb_mailbox.c']
-libusb = dependency('libusb-1.0')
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -1,12 +1,49 @@
...
-pkg.generate(libhoth_usb, name: 'libhoth_usb', description: 'Hoth USB interface library')
+pkg.generate(
+  hoth_usb_lib,
+  name: 'libhoth_usb',
+  description: 'Hoth USB interface library',
+  requires: hoth_usb_reqs)
 
-subdir('examples')
+if get_option('examples')
+  subdir('example')
```
korran, line 48: should be examples

wak-google: Done

meson.build
```diff
@@ -1,12 +1,49 @@
...
-libhoth_usb = library('hoth_usb', libhoth_usb_srcs, dependencies: libusb, install: true, \
-    version: meson.project_version())
 install_headers('libhoth_usb.h')
 
+hoth_usb_reqs = []
+foreach dep : hoth_usb_deps
+  if dep.type_name() == 'pkgconfig'
+    hoth_usb_reqs += dep
+  endif
+endforeach
```
eli-schwartz, line 38: I don't think you need this. Meson should automatically recurse into link dependencies of the libraries you built within this project, and add those as generated pkg-config dependencies as well.

cjevans-google:
Closed as obsolete: we were far too slow to review this properly, and the `meson.build` files have drifted since then.  Feel free to rework and open a new PR if this is still desireable.

PR #16, opened 2023-03-30
Add console snapshot logs

quadpixels:


examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+}
+
+static int command_payload_update(const struct htool_invocation* inv) {
+  struct libhoth_usb_device* dev = htool_libhoth_usb_device();
+  if (!dev) {
+    return -1;
+  }
+
+  bool dry_run = false;
+  char* x = getenv("DRY_RUN");
```
cjevans-google, line 520: Having dry-run be an environment variable is a little weird -- I think this would more naturally be a command-line flag? That might be inconvenient here, though.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+    RETRY:
+      if (!dry_run) {
+        result = htool_exec_hostcmd(
+            dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HOTH_PAYLOAD_UPDATE,
+            NULL, request1, sizeof(*request1) + request1->len, &response,
+            sizeof(response), &rlen);
+      }
+      if (result != 0) {
+        printf("Error, result=%d. will retry\n", result);
+        goto RETRY;
```
cjevans-google, line 605: This goto can be a loop:

```
while (1) {
  result = htool_exec_hostcmd(...);
  if (result == 0) {
    break;
  }
  printf("Error, ...\n");
}
```

But also, this retries indefinitely, and probably should give up eventually (after some number of tries, or a timeout).  Might be better to refactor this to a separate function, `htool_exec_hostcmd_with_retries` or similar.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+      offset1 += max_size_to_send;
+    }
+
+    idx++;
+  }
+
+  result = send_payload_update_command(dev, 2, 0, 0);  // Finalize
+  if (result != 0) {
+    printf("sending FINALIZE failed.\n");
+    return -1;
```
cjevans-google, line 624: Return here skips the `cleanup2` and `cleanup` blocks.  (I'm not especially fond of the `goto cleanup_label` pattern, since it often causes this sort of bug, but the `..._internal` function pattern is noisy.)

examples/htool_console.c
```diff
@@ -244,6 +244,40 @@ static int set_uart_config(struct libhoth_usb_device *dev,
...
+    int status = htool_exec_hostcmd(dev, EC_CMD_CONSOLE_REQUEST,
+      0, NULL, 0, NULL, 0, &response_bytes_written);
+    if (status != LIBHOTH_OK) {
+      printf("EC_CMD_CONSOLE_REQUEST not OK.\n");
+      return status;
+    }
+
+    struct ec_params_console_read_v1 read_request = { .subcmd = CONSOLE_READ_NEXT };
+    const size_t max_bytes_per_read = kMailboxSize - sizeof(struct ec_host_response);
+    char ret[12288] = { 0 };
```
cjevans-google, line 262: 12288 is a magic number ("max console buffer size"), and at least should be more clearly expressed as 0x3000.

examples/htool_console.c
```diff
@@ -244,6 +244,40 @@ static int set_uart_config(struct libhoth_usb_device *dev,
...
+    while (true) {
+      char buf[kMailboxSize];
+      status = htool_exec_hostcmd(
+        dev, EC_CMD_CONSOLE_READ,
+        0, &read_request, sizeof(read_request), buf, max_bytes_per_read, &response_bytes_written);
+      if (status != LIBHOTH_OK) {
+        printf("EC_CMD_CONSOLE_READ not OK.\n");
+        return status;
+      }
+      strcat(ret, buf);
```
cjevans-google, line 273: `strcat` will gleefully write past the end of the buffer, and there's no guarantee that the firmware only has 0x3000 bytes to provide.  At least change this to `strncat(ret, buf, sizeof(buf));`.

examples/host_commands.h
```diff
@@ -169,6 +169,8 @@ struct ec_request_reset_target {
 
 /* Reset the target device. */
 #define EC_PRV_CMD_HOTH_RESET_TARGET 0x0012
+#define EC_PRV_CMD_HOTH_PAYLOAD_UPDATE 0x0005
```
cjevans-google, line 172: nit: Include a blank line to separate from the reset-target above -- or maybe a comment `/* Payload update commands. */`

examples/host_commands.h
```diff
@@ -271,4 +273,16 @@ struct ec_channel_uart_config_set_req {
   struct ec_channel_uart_config config;
 } __attribute__((packed, aligned(4)));
 
+#define EC_CMD_CONSOLE_REQUEST 0x0097
+#define EC_CMD_CONSOLE_READ 0x0098
+
+enum ec_console_read_subcmd {
+	CONSOLE_READ_NEXT = 0,
+	CONSOLE_READ_RECENT
```
cjevans-google, line 281: nit: `CONSOLE_READ_RECENT = 1,`, as this is a protocol and the actual enum values must be stable.

(Yes, this is also missing in the header you consulted to add here, which we should also fix.)

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+  uint32_t version_major;
+  uint32_t version_minor;
+  uint32_t version_point;
+  uint32_t version_subpoint;
+  uint32_t descriptor_offset; /* can be used to pull the image hash/signature */
+} __attribute__((packed));
+
+const char* SpsEepromLockdownStatusString(uint8_t st) {
+  switch (st) {
+    case 0:
```
cjevans-google, line 278: nit: It's probably wise to have actual machine-readable enums for these values, here and below -- but I won't insist.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+    return -1;
+  }
+
+  uint8_t buffer[sizeof(struct payload_status_response_header) +
+                 2 * sizeof(struct payload_region_state)];
+  size_t rlen = 0;
+  int ret = htool_exec_hostcmd(
+      dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HOTH_PAYLOAD_STATUS, 0,
+      NULL, 0, &buffer, sizeof(buffer), &rlen);
+  printf("[do_payload_status] ret=%d, rlen=%d\n", ret, rlen);
```
cjevans-google, line 364: Stray debugging code?

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+  size_t rlen = 0;
+
+  int ret = htool_exec_hostcmd(
+      dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HOTH_PAYLOAD_UPDATE, 0,
+      &request, sizeof(request), buf, len, &rlen);
+  return ret;
+}
+
+int command_payload_read(struct libhoth_usb_device* dev, uint32_t offset, uint32_t len, uint8_t* buf) {
+  return send_payload_update_command_with_return_value(dev, 6, offset, len, buf);
```
cjevans-google, line 470: `6` is a magic number here -- the subcommands opcodes should be in a enum instead, especially if we're using them in multiple places.

(Same issue elsewhere, but I haven't noted every such case, to avoid comment spam.)

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+  const size_t buf_size = sizeof(struct payload_update_packet) + 2048;
+  uint8_t buffer[buf_size];
+
+  struct stat statbuf;
+  if (fstat(fd, &statbuf)) {
+    fprintf(stderr, "fstat error: %s\n", strerror(errno));
+    goto cleanup1;
+  }
+  if (statbuf.st_size > SIZE_MAX) {
+    fprintf(stderr, "file to large\n");
```
cjevans-google, line 560: typo: `s/to/too/`.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+    goto cleanup1;
+  }
+
+  size_t file_size = statbuf.st_size;
+  printf("File size: %zd\n", file_size);
+  uint8_t* file_data = mmap(NULL, file_size, PROT_READ, MAP_PRIVATE, fd, 0);
+
+  printf("2. Sending Data\n");
+  const size_t max_packet_size = 1024 - sizeof(struct payload_update_packet) -
```
cjevans-google, line 569: nit: `s/1024/kMailboxSize/` -- we have the constant, may as well use it.

(Also I'm intending to increase the mailbox size to 2KiB in the future.  Not relevant to this commit, but using the same constant will make that future change easier.)

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+            NULL, request1, sizeof(*request1) + request1->len, &response,
+            sizeof(response), &rlen);
+      }
+      if (result != 0) {
+        printf("Error, result=%d. will retry\n", result);
+        goto RETRY;
+      }
+      printf("[Frame %d] offset=%zd len=%zd, response=%u ", idx, offset1,
+             to_send, response);
+      for (int i = 0; i < 40; i++) {
```
cjevans-google, line 609: Stray debugging code? You probably don't need to print the first 40 bytes of each packet.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+}
+
+int command_payload_read(struct libhoth_usb_device* dev, uint32_t offset, uint32_t len, uint8_t* buf) {
+  return send_payload_update_command_with_return_value(dev, 6, offset, len, buf);
+}
+
+// Similar to Hoth::getPayloadSize
+static int command_payload_size(const struct htool_invocation* inv) {
+  printf("The payload_size program uses binary-search to determine the payload size.\n"
+         "Error message are expected. Just disregard the errors.\n");
```
cjevans-google, line 476: This is mildly suspicious: I'm not sure why a binary search is necessary here, and binary search is notoriously easy to implement incorrectly.  Seems like we're trying to figure out how large the backing eeprom is?  We might want to add this functionality to the underlying firmware directly.

Aside from that, if we know that the error messages here are expected, we ought to have a way to suppress them in the code itself (I presume some lower layer is printing them), rather than make the tool un-ergonomic to use.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+  printf("1. Sending Initiate\n");
+  if (!dry_run) {
+    result = send_payload_update_command(dev, 0, 0, 0);
+    if (result != 0) {
+      printf("sending INITIATE failed.\n");
+      return -1;
+    }
+  }
+
+  printf("source_file: %s\n", args.source_file);
```
cjevans-google, line 542: nit: There's a bunch of stray printf's throughout this code, which don't seem very necessary.  Progress output is worthwhile, but just echoing what the user asked for back to them isn't especially useful.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+
+  int fd = open(args.source_file, O_RDONLY, 0);
+  if (fd == -1) {
+    fprintf(stderr, "Error opening file %s: %s\n", args.source_file,
+            strerror(errno));
+    return -1;
+  }
+
+  const size_t buf_size = sizeof(struct payload_update_packet) + 2048;
```
cjevans-google, line 551: Is `2048` here arbitrary? ie, it just needs to be "big enough", but there's no reason to pick this specific number?  Based on code below, I think this can be just kMailboxSize?

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+      printf("\n");
+      data += max_size_to_send;
+      size -= max_size_to_send;
+      offset1 += max_size_to_send;
+    }
+
+    idx++;
+  }
+
+  result = send_payload_update_command(dev, 2, 0, 0);  // Finalize
```
cjevans-google, line 621: This code overlaps with `command_payload_finalize`; might be worthwhile to deduplicate.  Optional -- but if `payload update` always calls finalize, why do we have a separate command for `payload finalize`?

examples/htool_console.c
```diff
@@ -244,6 +244,40 @@ static int set_uart_config(struct libhoth_usb_device *dev,
 
 int htool_console_run(struct libhoth_usb_device *dev,
                       const struct htool_console_opts *opts) {
+  if (opts->snapshot) {
+    printf("Reading console snapshot\n");
+    // 1. Create Debug interface (should be already created by this time.)
+    // 2. read console snapshot
+    const size_t kMailboxSize = 1024;
```
cjevans-google, line 251: This constant is generally useful, and should be in a more-widely-visible header file (so you can use it in the other locations I've noted in this review.)

examples/htool_console.c
```diff
@@ -244,6 +244,40 @@ static int set_uart_config(struct libhoth_usb_device *dev,
 
 int htool_console_run(struct libhoth_usb_device *dev,
                       const struct htool_console_opts *opts) {
+  if (opts->snapshot) {
```
cjevans-google, line 247: nit: If snapshot has so little overlap with the existing `htool_console_run` implementation, it may as well be its own function, `htool_console_snapshot`.  (Not necessary to have a different user-visible command, just nitpicking on code structure).

cjevans-google (changes requested):
Thanks for contributing this improvement -- we have a lot of work I want to get done in libhoth (time permitting, sigh), and this at least unblocks the most immediate needs.  Apologies that I'm being a bit picky in this review.

examples/host_commands.h
quadpixels: Removed payload-related constants from this change (did not intend to add them)

examples/htool.c
quadpixels: Removed payload-related structures and constants from this change for now.

examples/htool.c
quadpixels: Removed stray debugging code.

examples/htool.c
```diff
@@ -251,6 +251,413 @@ static int do_target_reset(uint32_t reset_option) {
...
+    printf("  key_index:  %u\n", rs->key_index);
+    printf("  image_family: (0x%08x)\n", rs->image_family);
+    printf("  version: %u.%u.%u.%u\n", rs->version_major, rs->version_minor,
+           rs->version_point, rs->version_subpoint);
+    printf("  descriptor_offset: 0x%08x\n", rs->descriptor_offset);
+  }
+  return 0;
+}
+
+static int command_payload_version(const struct htool_invocation* inv) {
```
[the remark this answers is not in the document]
quadpixels: It is removed, since the "payload version" is just a subset of "payload status". But it doesn't matter that much because all payload-related code is removed from this change for now and will be created in a separate change.

examples/htool.c
quadpixels: Sure, will remove payload-related code for now, and address them in a separate change.

examples/htool.c
quadpixels: Yes, it is trying to figure out the size of the EEPROM by finding out the largest address that could be read. The binary-search algorithm is done by the BMC-side daemon.

examples/htool.c
quadpixels: It is a debugging statement, and will be removed.

examples/htool.c
quadpixels: Removed stray print statements.

examples/htool.c
quadpixels: Yes, it is intended to be the mailbox size. Will address it in a separate change focused on payload update.

examples/htool.c
quadpixels: Done.

examples/htool.c
quadpixels: Acknowledged.

examples/htool.c
quadpixels: Thanks for the comments, will address this in a separate change focused on payload update.

examples/htool.c
quadpixels: Because I extracted "payload finalize" into a separate function while debugging it, will de-duplicate.

examples/htool_console.c
quadpixels: Moved to a separate function `htool_console_snapshot`.

examples/htool_console.c
quadpixels: Done.

examples/htool_console.c
quadpixels: Done.

examples/htool_cmd.c
```diff
@@ -177,17 +177,21 @@ int htool_get_param_string(const struct htool_invocation* inv, const char* name,
 
 int htool_get_param_bool(const struct htool_invocation* inv, const char* name,
                          bool* value) {
+  printf(">> htool_get_param_bool\n");
```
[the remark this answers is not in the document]
quadpixels: Removed stray debugging code.

examples/htool_console.c
quadpixels: Moved to "host_commands.h".

examples/host_commands.h
quadpixels: Done.
It seems this file already contains contents prefixed with `EC_` that are copied-and-pasted from the header file I consulted to (ec_header.h), so I assumed "copying what is needed to this file" is the usual practice.

cjevans-google (approved):

cjevans-google (approved):

PR #18, opened 2023-04-19
Build improvements

gyandreev:
Add warnings as errors flag to meson and blaze CI file.
Fix some warnings from console snapshot change.

### DIFF
--- a/examples/htool_console.c
+++ b/examples/htool_console.c
@@ -320,7 +320,7 @@ int htool_console_snapshot(struct libhoth_device *dev) {
       fprintf(stderr, "EC_CMD_CONSOLE_READ status: %d\n", status);
       return status;
     }
-    printf("%s", buf);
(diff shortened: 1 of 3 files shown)


examples/htool_console.c
```diff
@@ -322,9 +320,9 @@ int htool_console_snapshot(struct libhoth_device *dev) {
       fprintf(stderr, "EC_CMD_CONSOLE_READ status: %d\n", status);
       return status;
     }
-    strncat(ret, buf, sizeof(buf));
+    printf("%s", buf);
```
korran, line 323: If haven doesn't null-terminate the snapshot, this could print out garbage from the stack.

I think you want something like

```
fwrite(buf, strnlen(buf, sizeof(buf)), 1, stdout);
```

gyandreev: Ack. Fixed

korran (approved):

PR #19, opened 2023-04-26
Add --noenter4b flag for SPI update

aranikam:
For spi operations for targets that don't support 4B ……mode

### DIFF
--- a/examples/htool.c
+++ b/examples/htool.c
@@ -104,15 +104,38 @@ static int force_write(int fd, const void* buf, size_t size) {
   return 0;
 }
 
+static int get_address_mode( const char* address_mode, bool* is_4_byte,
+bool* enter_4byte) {
+  if (!strcmp(address_mode, "3B/4B")) {
+    *is_4_byte = true;
+    *enter_4byte = true;
+  }
+  else if (!strcmp(address_mode, "3B")) {
+    *is_4_byte = false;
+    *enter_4byte = false;
+  }
+  else if(!strcmp(address_mode, "4B")) {
+    *is_4_byte = true;
+    *enter_4byte = false;
+  }
+  else {
+    fprintf(stderr, "Invalid address_mode value: %s\n", address_mode);
+    return -1;
+  }
+  return 0;
+}
+
 static int command_spi_read(const struct htool_invocation* inv) {
   struct {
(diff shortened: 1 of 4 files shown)


examples/htool.c
```diff
@@ -422,6 +426,8 @@ static const struct htool_cmd CMDS[] = {
                 {HTOOL_FLAG_VALUE, 'n', "length",
                  .desc = "the number of bytes to read"},
                 {HTOOL_POSITIONAL, .name = "dest-file"},
+                {HTOOL_FLAG_BOOL, 'b', "noenter4b", "false",
```
gyandreev, line 429: Maybe we can change the character to something else and reserve b for binary?

examples/htool_spi_proxy.h
```diff
@@ -25,7 +25,7 @@ struct htool_spi_proxy {
   bool is_4_byte;
 };
 
-int htool_spi_proxy_init(struct htool_spi_proxy* spi, struct libhoth_device* dev);
+int htool_spi_proxy_init(struct htool_spi_proxy* spi, struct libhoth_device* dev, bool noenter4b);
```
gyandreev, line 28: Variable name in definition does not match declaration. Is that intended for documentation reasons?

gyandreev: Actually, they are contradictory. `noenter4b` != `is_4_byte`

daimeng-wang:
We should also consider situations where the Hoth is 4B mode only and doesn't support Enter/Exit4B

examples/htool.c
```diff
@@ -433,6 +474,10 @@ static const struct htool_cmd CMDS[] = {
                 {HTOOL_FLAG_VALUE, 's', "start", "0", .desc = "start address"},
                 {HTOOL_FLAG_BOOL, 'v', "verify", "true"},
                 {HTOOL_POSITIONAL, .name = "source-file"},
+                {HTOOL_FLAG_VALUE, 'a', "address_mode", "3B/4B",
```
gyandreev, line 477: Nit suggestion, can we move the FLAG_VALUE to before the POSITIONAL? Just for readability so the positional param does not get buried behind other flags.

cjevans-google (approved):

PR #20, opened 2023-05-02
Add payload status

quadpixels:


examples/htool_payload.h
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBHOTH_EXAMPLES_HTOOL_PAYLOAD_H_
+#define LIBHOTH_EXAMPLES_HTOOL_PAYLOAD_H_
+
+#include <stdint.h>
+
+struct payload_status_response_header {
```
cjevans-google, line 20: Should these two structure definitions be in host_commands.h instead?

cjevans-google (approved):

examples/htool_payload.h
quadpixels: Yes this sounds reasonable, since `payload_status_response_header` is directly bit-cast from the response buffer, similarly to other structs in host_commands.h. So moved to `host_commands.h`.

PR #32, opened 2023-05-11
Implement authorization record host command

daimeng-wang:


daimeng-wang:
Internal review comment:

* change in mtd_read to allow zero-length reads as a no-op should be in a separate commit, I think -- not totally sure why this is needed
A: This is to avoid error when resp.data_len = 0, which happens in authz_record set

daimeng-wang:
* authorization_record_from_hex_string doesn't appear to handle invalid characters, or strtoul errors? also not sure about the two-byte memcpy there
A: Fixed. strtoul only throws E_RANGE error which should be impossible here.

gyandreev (approved):

gyandreev (approved):

PR #34, opened 2023-06-29
Update print_version_header.sh

nnooney:
Change the sed flag from -E to -r. 

-E is on some versions of sed for compatibility with BSD; -r is the GNU variant. See https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed for details.


nnooney:
Removing as I found a different workaround for building.

PR #35, opened 2023-06-29
Add default_applicable_licenses to BUILD rules

nnooney:


cjevans-google:
Handled this issue internally: default_applicable_licesnses is a Google-internal constraint, and doesn't affect the public version.

PR #39, opened 2023-12-11
Add BUILD rules to support using libhoth with external tools

nnooney:
- Expose USB and SPI devices as separate build targets
- Add library targets (header and lib) for API use


BUILD
```diff
@@ -2,23 +2,8 @@ package(default_visibility = ["//visibility:public"])
...
     name = "libhoth",
-    srcs = [
-        "libhoth.c",
-        "libhoth_mtd.c",
-        "libhoth_spi.c",
-        "libhoth_usb.c",
-        "libhoth_usb_fifo.c",
-        "libhoth_usb_mailbox.c",
-    ],
-    hdrs = [
```
cjevans-google, line 13: This change (removing headers from `:libhoth`) might break downstream integrations of libhoth, who will have to include their desired subset of {`:libhoth_mtd`, `:libhoth_spi`, `:libhoth_usb`} for the backends they need.  This is probably necessary eventually, but I this we want to have this change include `deps = ["//:libhoth_mtd", "//:libhoth_spi", "//:libhoth_usb"],` here to avoid this churn for the moment.

cjevans-google (changes requested):

BUILD
nnooney: Added deps on `:libhoth_mtd`, `:libhoth_spi`, and `:libhoth_usb`.

nnooney: Update: removed these deps, as it would introduce a cyclic build dependency.

cjevans-google (approved):

PR #42, opened 2024-01-05
htool: Allow console snapshot on proxy channels

daimeng-wang:
`-h` flag with UART proxy still creates a live console, which is undesired sometimes.

Allow `-s` flag to work with `-c` flag. When both are set, will perform console read until the current buffer pointer.

### DIFF
--- a/examples/htool_console.c
+++ b/examples/htool_console.c
@@ -359,7 +359,7 @@ int htool_console_run(struct libhoth_device *dev,
   return status;
 }
 
-int htool_console_snapshot(struct libhoth_device *dev) {
+int htool_console_snapshot_legacy(struct libhoth_device *dev) {
   size_t response_bytes_written;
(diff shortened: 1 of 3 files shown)


examples/htool_console.c
```diff
@@ -386,3 +386,33 @@ int htool_console_snapshot(struct libhoth_device *dev) {
...
 }
+
+int htool_console_snapshot(struct libhoth_device *dev,
+                           const struct htool_console_opts *opts) {
+  // Legacy host commands for console snapshot.
+  if (!opts->channel_id) {
+    return htool_console_snapshot_legacy(dev);
+  }
+  // Modern host commands similar to htool_console_run.
+  // Starting from current_offset - 0x80000000. Repeat read until reaching
```
gyandreev, line 397: Maybe add a comment as to why you are using `0x80000000` here?

gyandreev (approved):

PR #43, opened 2024-01-10
Update error messages in htool_exec_hostcmd

nnooney:
Original logs referenced different function names than are actually invoked, making it harder to trace the program execution.


daimeng-wang (approved):
LGTM. Thanks!

PR #44, opened 2024-01-31
Add API to call SRTM host command.

Quanzu:


examples/host_commands.h
```diff
@@ -581,4 +581,15 @@ struct ec_response_persistent_panic_info {
   int32_t persistent_panic_record_version;
 } __attribute__((packed));
 
+// This command allows callers to push initial measurements into PCR0. This
+// command will fail if the TPM has already been initialized, or if the
```
jhand2, line 585: nit:
```suggestion
// command will fail if the TPM has already been started up, or if the
```

Init and startup are different things in TPM fw

examples/htool.c
```diff
@@ -536,6 +536,36 @@ static int command_passthrough_enable(const struct htool_invocation* inv) {
...
+  if (!dev) {
+    return -1;
+  }
+
+  const char* measurement;
+  if (htool_get_param_string(inv, "measurement", &measurement) != 0) {
+    return -1;
+  }
+
+  size_t measurement_length = strlen(measurement);
```
jhand2, line 550: I don't think it will work to pass raw binary data here. (a) it makes to command pretty hard to use and (b) if the data contains any null bytes, the strlen check will terminate early.

I think there are a couple ways to do this that might make sense:

1. Pass a file with binary data that htool can read
2. Pass a hex string which we decode to binary

1 seems a little easier to implement, but 2 seems easier to use.

jhand2 (approved):

jhand2 (changes requested):
Whoops, didn't mean to approve

jhand2 (approved):

examples/host_commands.h
```diff
@@ -581,4 +581,15 @@ struct ec_response_persistent_panic_info {
   int32_t persistent_panic_record_version;
 } __attribute__((packed));
 
+// This command allows callers to push initial measurements into PCR0. This
+// command will fail if the TPM has already been started up, or if the
+// data to measure exceeds SRTM_DATA_MAX_SIZE_BYTES.
+#define EC_PRV_CMD_HAVEN_SRTM 0x0044
```
rkr35, line 587: Replace `HAVEN` with `HOTH` to be consistent with existing commands.

rkr35 (approved):

PR #45, opened 2024-02-22
Uart break

gyandreev:
Add the uart break command for the hoth uart proxy.
With console open, break can be sent by <CTRL + T, b>

### DIFF
--- a/examples/htool_console.c
+++ b/examples/htool_console.c
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


examples/htool_console.c
```diff
@@ -150,7 +150,12 @@ void restore_terminal(int fd, const struct termios *old_termios) {
 // in-place escape sequence processing.
 // looks for ctrl-T/q.
```
rkr35, line 151: nit: generalize this comment to say that it's looking for <kbd>CTRL</kbd> + <kbd>T</kbd> + `escape char`.

rkr35 (approved):

examples/htool_console.c
gyandreev: Done

PR #47, opened 2024-02-23
htool: don't print help msg when channel param is optional and missing

rkr35:
`htool console` supports these flags:
```
  -c --channel
        Which channel to talk to. Typically a fourcc code.
  -s --snapshot (default: "false")
        Print a snapshot of most recent console messages.
```
Those flags can be combined. Here's the intended behavior:

|             |            **no `-s`**           |        **`-s`**        |
|-------------|:--------------------------------:|:----------------------:|
| **no `-c`** | print help msg: `-c` is required | print Hoth console log |
|   **`-c`**  | open live UART console           | print UART console log |

There's a bug with the "no `-c`" + "`-s`" case: htool emits an error message saying the channel is missing.

```
$ htool console -s
Unspecified required flag: --channel (or -c)
Usage: console
...snip usage text...
hoth console log
```

As described in the above table, the channel is not required for this case. This PR fixes this bug.

### DIFF
--- a/examples/htool.c
+++ b/examples/htool.c
@@ -458,8 +458,12 @@ static int command_console(const struct htool_invocation* inv) {
       return -1;
     }
   } else {
-    // Channel is optional in snapshot mode: 0 = legacy host command.
-    if (htool_get_param_u32_or_fourcc(inv, "channel", &opts.channel_id)) {
+    if (htool_has_param(inv, "channel")) {
+      if (htool_get_param_u32_or_fourcc(inv, "channel", &opts.channel_id)) {
+        return -1;
+      }
+    } else {
+      // Channel is optional in snapshot mode: 0 = legacy host command.
       opts.channel_id = 0;
     }
   }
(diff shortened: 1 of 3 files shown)


anthony-huang (approved):

gyandreev:
This is already fixed by: https://github.com/google/libhoth/pull/21

gyandreev:
Weird, it is broken again

gyandreev:
Looks like this may be the culprit: https://github.com/google/libhoth/pull/42

rkr35:
> This is already fixed by: #21

I believe that fix is related but distinct. This PR fixes a bug we recently introduced in this chunk of code: https://github.com/google/libhoth/commit/e4827163741e0804f12ac96c81b8e97649be6795#diff-c2a11f3cffdb0069f161a066a2bdba961967a7411de749ceb25788948d3e676fR459.

The call to `htool_get_param_u32_or_fourcc` will print an error message if `--channel` is absent, which is undesirable if we want the Hoth snapshot instead of the UART snapshot.

gyandreev:
ACK Yeah. That's because `htool_get_param_u32_or_fourcc` is using param_required.

gyandreev (approved):

PR #48, opened 2024-03-06
I2c commands

gyandreev:


examples/htool_i2c.c
```diff
@@ -0,0 +1,220 @@
...
+  request.bus_number = (uint8_t)(bus & 0xFF);
+  request.dev_address = (uint8_t)(addr & 0x7F);
+  request.speed_khz = (uint16_t)(freq & 0xFFFF);
+  request.flags = (no_stop ? I2C_BITS_NO_STOP : 0) | I2C_BITS_WRITE;
+  request.size_read = 0;
+
+  uint16_t idx = 0;
+  char *tk = strtok(byte_stream, " ");
+  while (tk && (idx < I2C_TRANSFER_DATA_MAX_SIZE_BYTES)) {
+    unsigned long int parsed = strtoul(tk, NULL, 0);
```
rkr35, line 166: `strtoul` will silently return `0` if it can't perform the conversion. We should check for and report this failure.

examples/htool_i2c.h
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBHOTH_EXAMPLES_HTOOL_I2C_H_
+#define LIBHOTH_EXAMPLES_HTOOL_I2C_H_
+
+#include <stdint.h>
```
rkr35, line 18: nit: we probably don't need to include this header.

examples/htool_i2c.c
```diff
@@ -0,0 +1,220 @@
...
+            sizeof(response), rLen);
+    return -1;
+  }
+
+  struct ec_response_i2c_transfer *pI2t =
+      (struct ec_response_i2c_transfer *)(response);
+
+  printf("Read %u bytes from 0x%02x\n", pI2t->read_bytes, request.dev_address);
+  for (uint16_t i = 0; i < pI2t->read_bytes; i++) {
+    printf("0x%02X ", pI2t->resp_bytes[i]);
```
rkr35, line 129: Same "space-separated vs. flattened byte string" comment as elsewhere.

examples/htool_i2c.c
```diff
@@ -0,0 +1,220 @@
...
+
+  struct ec_request_i2c_transfer request;
+  request.bus_number = (uint8_t)(bus & 0xFF);
+  request.dev_address = (uint8_t)(addr & 0x7F);
+  request.speed_khz = (uint16_t)(freq & 0xFFFF);
+  request.flags = (no_stop ? I2C_BITS_NO_STOP : 0) | I2C_BITS_WRITE;
+  request.size_read = 0;
+
+  uint16_t idx = 0;
+  char *tk = strtok(byte_stream, " ");
```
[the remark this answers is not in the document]
gyandreev: No particular reason except it is easier to read when running these transactions manually/for debugging as this is mostly intended to

gyandreev:
Added responses

examples/htool_i2c.c
rkr35: Right, we probably want to do this or a similar check. If these commands are going to be used manually, then someone may accidentally type an invalid byte that `strtoul` silently converts to a zero.

examples/htool_i2c.c
rkr35: Ack.

examples/htool_i2c.c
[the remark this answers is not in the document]
rkr35: Ack.

examples/htool_i2c.c
gyandreev: Checking the output pointer now as well as errno

examples/htool_i2c.c
```diff
@@ -0,0 +1,226 @@
...
+  while (tk && (idx < I2C_TRANSFER_DATA_MAX_SIZE_BYTES)) {
+    char *endptr;
+    unsigned long int parsed = strtoul(tk, &endptr, 0);
+    if ((errno != 0 && parsed == 0) || tk == endptr) {
```
rkr35, line 169: According to the documentation for [`strtoul`](https://en.cppreference.com/w/c/string/byte/strtoul), the function only sets `errno`  when the converted integer is outside the range of `unsigned long`, and in that case, `parsed` would be `ULONG_MAX` instead of `0`. The `tk == endptr` already checks if the conversion is valid.

I would include two more checks:

1. Check that `parsed` is actually a byte. `"0x123"` should fail.
2. Check that we parsed the entire input. `"0x"` and `"1a"` should fail.

```suggestion
    if (tk == endptr || parsed > UINT8_MAX || *endptr != '\0') {
```

gyandreev: ACK. I misunderstood the way the documentation was worded; errno is set and the result is ULONG_MAX or 0 depending on if a conversion can be completed or not. The documentation for `strtol` is clearer.

I have applied your suggested change (and fixed an array index error I saw after proofing)

rkr35 (approved):

examples/htool_i2c.c
rkr35: Thanks!

PR #52, opened 2024-04-08
Fix formatting of authorization record.

cjevans-google:
Some systems consider `char` to be signed by default, which causes the system implementation of `printf("%02x", (char)0xd6);` to print `ffffffd6`, instead of just `d6`.  As the nonce almost always has "negative" bytes, this badly corrupts the record formatting.

Also add a _Static_assert for the exact size of the authorization record, for good measure.


rkr35 (approved):
LGTM

PR #59, opened 2024-05-10
Hex dump command response on checksum failure

rkr35:
The dump helps us debug why the response checksum is bad.

Example sending `EC_CMD_HELLO` with a corrupted response body:
```
$ echo -ne '\x03\x18\x01\x00\x00\x00\x04\x00\xd0\xc0\xb0\xa0' | htool raw_host_command
Error: response checksum (255) != 0
Response header:
0x0000: 03 0f 00 00 04 00 00 00                          |........        |
Response body:
0x0000: d5 c3 b2 a1                                      |....            |
EC response header invalid: -22
```

### DIFF
--- a/examples/ec_util.c
+++ b/examples/ec_util.c
@@ -18,6 +18,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <ctype.h>
 
 uint8_t calculate_ec_command_checksum(const void* header, size_t header_size,
                                       const void* data, size_t data_size) {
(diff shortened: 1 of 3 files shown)


send.bash
cjevans-google, line 1: Did you mean to include this file?

examples/ec_util.c
```diff
@@ -18,6 +18,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <ctype.h>
```
cjevans-google: nit: sort includes

cjevans-google (changes requested):

send.bash
rkr35: Removed.

examples/ec_util.c
rkr35: Done.

cjevans-google (approved):

PR #60, opened 2024-05-13
Add --spidev_atomic.

cjevans-google:
Some systems have special-purpose spidev drivers that can only enforce atomicity of the mailbox request if the entire SPI transaction sequence (write followed by read) is given as a single atomic `ioctl`.  Add the --spidev_atomic flag to enable this special-case behavior.

### DIFF
--- a/libhoth_spi.c
+++ b/libhoth_spi.c
@@ -17,6 +17,7 @@
 #include <fcntl.h>
 #include <linux/spi/spidev.h>
 #include <linux/types.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -30,7 +31,10 @@
 struct libhoth_spi_device {
   int fd;
   unsigned int mailbox_address;
-  unsigned int address_mode_4b;
+  bool address_mode_4b;
+
+  void* buffered_request;
+  size_t buffered_request_size;
 };
 
 int libhoth_spi_send_request(struct libhoth_device* dev, const void* request,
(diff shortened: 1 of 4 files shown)


libhoth_spi.c
```diff
@@ -195,10 +202,15 @@ int libhoth_spi_open(const struct libhoth_spi_device_init_options* options,
 
   spi_dev->fd = fd;
   spi_dev->mailbox_address = options->mailbox;
-  spi_dev->address_mode_4b = 1;
+  spi_dev->address_mode_4b = true;
```
rkr35, line 205: nit: if we're always assuming 4B, it might be worthwhile to remove and simplify the function signatures.

Not required for this PR since it's cleanup I can pick up.

libhoth_spi.c
```diff
@@ -284,6 +296,111 @@ int libhoth_spi_receive_response(struct libhoth_device* dev, void* response,
...
+
+int libhoth_spi_send_and_receive_response(struct libhoth_device* dev,
+                                          void* response,
+                                          size_t max_response_size,
+                                          size_t* actual_size, int timeout_ms) {
+  if (dev == NULL) {
+    return LIBHOTH_ERR_INVALID_PARAMETER;
+  }
+
+  if (max_response_size < 8) {
```
rkr35, line 327: nit: replace `8` with `sizeof(struct ec_host_response)`

libhoth_spi.c
```diff
@@ -284,6 +296,111 @@ int libhoth_spi_receive_response(struct libhoth_device* dev, void* response,
...
+
+  int rc = LIBHOTH_OK;
+  int status = ioctl(spi_dev->fd, SPI_IOC_MESSAGE(5), xfer);
+  if (status < 0) {
+    rc = LIBHOTH_ERR_FAIL;
+  } else {
+    if (actual_size) {
+      struct ec_host_response* host_response =
+          (struct ec_host_response*)response;
+      *actual_size = (size_t)host_response->data_len + 8;
```
rkr35, line 393: nit: and the `8` here, too.

rkr35 (approved):

libhoth_spi.c
cjevans-google: Done, here and elsewhere.

libhoth_spi.c
cjevans-google: Done, here and elsewhere.

libhoth_spi.c
cjevans-google: Actually, the opposite: the user ought to be able to enable this partially-implemented feature; it's merely coincidental that all our current uses of the spidev transport use 4-byte address mode exclusively.

PR #63, opened 2024-05-17
Fix formatting for `uint64_t` statistics fields

paipeng-g:
"time_since_hoth_boot_us" is uint64 and should be printed using "%llu" instead of "%lu".

### DIFF
--- a/examples/htool_statistics.c
+++ b/examples/htool_statistics.c
@@ -215,7 +215,8 @@ int htool_statistics() {
   if (flags & kResetFlagApWatchdog) printf(kIndent, "ResetFlagApWatchdog");
   // clang-format on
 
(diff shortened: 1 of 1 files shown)


rkr35:
This implementation suggests a discrepancy in how `uint64_t` is defined for the target architecture.

If `uint64_t` is  `unsigned long int`, then `%lu` is fine.
If `uint64_t` is  `unsigned long long int`, then `%llu` is required.

Perhaps better to use [`PRIu64`](https://en.cppreference.com/w/c/types/integer#Format_constants_for_the_fprintf_family_of_functions) to pick the correct format string.

Something like:
```c
uint64_t foo = UINT64_MAX;
printf("foo = %" PRIu64 "\n", foo);
```

rkr35 (approved):

PR #65, opened 2024-06-26
Add support for reading IDCODE and testing BYPASS via JTAG

xorptr:
Add support for reading IDCODE and testing BYPASS via JTAG from a device connected to target

### DIFF
--- a/examples/htool_jtag.c
+++ b/examples/htool_jtag.c
@@ -1,0 +1,207 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "htool_jtag.h"
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "host_commands.h"
+#include "htool.h"
+#include "htool_cmd.h"
+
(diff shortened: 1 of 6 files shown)


examples/htool_jtag.c
```diff
@@ -0,0 +1,199 @@
...
+  printf("IDCODE: 0x%08x\n", response.idcode);
+  return 0;
+}
+
+// Parse exactly `expected_bytes_count` byte tokens from `param_value` string
+// into `byte_sequence`.
+// If fewer than `expected_bytes_count` byte tokens are present in
+// `param_value`, return code indicates error.
+// If more than `expected_bytes_count` byte tokens are present in `param_value`,
+// extra bytes are ignored and no error is indicated
```
xorptr, line 81: I am assuming this behavior is okay. If an error needs to be returned, I can change the code here.

examples/htool_jtag.c
```diff
@@ -0,0 +1,199 @@
...
+      .operation =
+          {
+              .clk_idiv = (uint16_t)clk_idiv,
+              .operation = EC_JTAG_OP_TEST_BYPASS,
+          },
+  };
+
+  int ret = parse_string_param_into_byte_sequence(
+      tdi_bytes, request.params.tdi_pattern, EC_JTAG_TEST_BYPASS_PATTERN_LEN);
+  if (ret < 0) {
```
xorptr, line 141: Change to `ret != 0`

examples/htool_jtag.c
rkr35: Someone trying to send a pattern greater than 64 bytes might be confused why htool is truncating their input, in which case a printed warning would clarify matters. Likewise, someone trying to send exactly 64 bytes should be made aware when their input is unintentionally longer.

examples/htool_jtag.c
rkr35: Agreed

examples/htool_jtag.c
xorptr: Fixed. The new changes should have the function accept exactly `expected_bytes_count` bytes in form of string tokens (no more and no less)

examples/htool_jtag.c
xorptr: Fixed

rkr35 (approved):

PR #67, opened 2024-06-28
Add `authz_host_command build/send`

rkr35:
Adds two commands:
1. `htool authz_host_command build OPCODE`
2. `htool authz_host_command send COMMAND_BYTES`

Example usage:
```
$ htool authz_host_command build 1
0000000000000000...

$ htool authz_host_command send cdf93af5308e4a71...
```

### DIFF
--- a/examples/htool_authz_command.c
+++ b/examples/htool_authz_command.c
@@ -1,0 +1,95 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "host_commands.h"
+
+struct ec_authorized_command_request authz_command_build_request(
+    uint64_t hardware_identity, uint32_t opcode, uint32_t key_info,
+    const uint32_t* nonce) {
+  struct ec_authorized_command_request request = {
+      .version = AUTHORIZED_COMMAND_VERSION,
+      .size = sizeof(request),
+      .key_info = key_info,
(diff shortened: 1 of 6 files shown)


examples/htool_authz_command.c
```diff
@@ -0,0 +1,81 @@
...
+    return -1;
+  }
+
+  uint8_t* out = (uint8_t*)request;
+
+  for (int i = 0; i < sizeof(*request); i++) {
+    uint8_t nibble_0 = parse_nibble(hexstring[2 * i]);
+    uint8_t nibble_1 = parse_nibble(hexstring[2 * i + 1]);
+    if (nibble_0 == UINT8_MAX || nibble_1 == UINT8_MAX) {
+      fprintf(stderr, "Invalid hex character at byte %d\n", i);
```
daimeng-wang, line 74: Do we want the same error output if hex string is longer than the struct size?

examples/htool_authz_command.c
```diff
@@ -0,0 +1,81 @@
...
+  if (actual_len != expected_len) {
+    fprintf(stderr, "Bad hexstring length. Got %ld but expected %ld\n",
+            actual_len, expected_len);
+    return -1;
+  }
+
+  uint8_t* out = (uint8_t*)request;
+
+  for (int i = 0; i < sizeof(*request); i++) {
+    uint8_t nibble_0 = parse_nibble(hexstring[2 * i]);
```
daimeng-wang, line 71: Can we use strtol() here?

examples/htool_authz_command.c
```diff
@@ -0,0 +1,81 @@
...
+    return c - 'A' + 0xa;
+  }
+  return UINT8_MAX;
+}
+
+int authz_command_hex_to_struct(const char* hexstring,
+                                struct ec_authorized_command_request* request) {
+  size_t actual_len = strlen(hexstring);
+  size_t expected_len = 2 * sizeof(*request);
+  if (actual_len != expected_len) {
```
daimeng-wang, line 62: Do we want the same output when provided hex string is longer than the struct?

daimeng-wang (commented):
Generally LGTM. See in-code comments

examples/htool_authz_command.c
rkr35: Added additional error messages to clarify that command payloads are unsupported.

examples/htool_authz_command.c
rkr35: We could, but I find the error-handling and state tracking more cumbersome, and `strtoul()`  accepts more characters than we care about, e.g., `+`, `-`, `0x`.

daimeng-wang (approved):

PR #84, opened 2025-02-14
Move progress to protocol/

stevenportley:


cjevans-google (changes requested):
Does it really makes sense for the progress bar support to be in `protocol/`?  The underlying library `libhoth` wants the means to provide progress, but it's up to the calling code to decide how to implement that.  In `htool`, it's sensible to print to the terminal, but that won't make sense in other systems using `protocol/` (which might prefer to ignore the progress indicator, or direct its output elsewhere).

stevenportley:
The protocols that want to provide some form of progress report need some sort of callback definition.  I don't see why that can't just be `libhoth_progress` and allow callers to provide their own implementation of it. 

If you're referring to `libhoth_progress_stderr_func` being in `protocol/`, I've just been thinking about it as a default option that many callers will want to re-use.

cjevans-google:
Ok, that argument makes sense to me.  I could still quibble that this might be clearer in a separate module (or to put `libhoth_progress_stderr_func` elsewhere), but I won't object.

cjevans-google (approved):

PR #85, opened 2025-02-19
Add payload inspection to htool/libhoth

stevenportley:
Add payload inspection to libHoth and an option to htool:

htool payload info protocol/test/test_payload.bin 
Payload Info:
  name: test layout                     
  family: 2
  version: 1.0.0.0
  type: 0
  hash: 50316da1d3b06ff87989aa8bcd48a33ecc4bfe2114ded138ee594f697d58b3

### DIFF
--- a/examples/htool_payload.c
+++ b/examples/htool_payload.c
@@ -14,11 +14,19 @@
 
 #include "htool_payload.h"
 
+#include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 #include "host_commands.h"
 #include "htool.h"
+#include "protocol/payload_info.h"
 #include "protocol/payload_status.h"
 
 int htool_payload_status() {
@@ -66,5 +74,71 @@ int htool_payload_status() {
            rs->version_point, rs->version_subpoint);
     printf("  descriptor_offset: 0x%08x\n", rs->descriptor_offset);
   }
+  return 0;
+}
+
+int htool_payload_info(const struct htool_invocation* inv) {
+  struct libhoth_device* dev = htool_libhoth_device();
+  if (!dev) {
+    return -1;
+  }
+
+  const char* image_file;
+  if (htool_get_param_string(inv, "source-file", &image_file) != 0) {
+    return -1;
+  }
+
+  int fd = open(image_file, O_RDONLY, 0);
+  if (fd == -1) {
+    fprintf(stderr, "Error opening file %s: %s\n", image_file, strerror(errno));
+    return -1;
+  }
+  struct stat statbuf;
+  if (fstat(fd, &statbuf)) {
+    fprintf(stderr, "fstat error: %s\n", strerror(errno));
+    goto cleanup2;
+  }
+  if (statbuf.st_size > SIZE_MAX) {
+    fprintf(stderr, "file too large\n");
+    goto cleanup2;
+  }
+
+  uint8_t* image = mmap(NULL, statbuf.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+  if (image == MAP_FAILED) {
+    fprintf(stderr, "mmap error: %s\n", strerror(errno));
+    goto cleanup2;
+  }
+
+  struct payload_info info;
+  if (!libhoth_payload_info(image, statbuf.st_size, &info)) {
+    fprintf(stderr, "Failed to parse payload image.  Is this a titan image?\n");
+    goto cleanup;
+  }
+
+  printf("Payload Info:\n");
+  printf("  name: %-32s\n", info.image_name);
+  printf("  family: %u\n", info.image_family);
+  printf("  version: %u.%u.%u.%u\n", info.image_version.major,
+         info.image_version.minor, info.image_version.point,
+         info.image_version.subpoint);
+  printf("  type: %u\n", info.image_type);
+  printf("  hash: ");
+  for (int i = 0; i < sizeof(info.image_hash); i++) {
+    printf("%02x", info.image_hash[i]);
+  }
+  printf("\n");
+
+cleanup:
+  if(munmap(image, statbuf.st_size) != 0) {
+    fprintf(stderr, "munmap error: %s\n", strerror(errno));
+    return -1;
+  }
+
+cleanup2:
+  if (close(fd) != 0) {
+    fprintf(stderr, "close error: %s\n", strerror(errno));
+    return -1;
+  }
+
   return 0;
 }
--- a/protocol/payload_info.c
+++ b/protocol/payload_info.c
@@ -1,0 +1,68 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "payload_info.h"
+
+#include <string.h>
+
+const struct image_descriptor* libhoth_find_image_descriptor(const uint8_t* image,
+                                                       size_t len) {
+  for (size_t off = 0; off + sizeof(struct image_descriptor) - 1 < len;
+       off += TITAN_IMAGE_DESCRIPTOR_ALIGNMENT) {
+    int64_t magic_candidate;
+    memcpy(&magic_candidate, image + off, sizeof(magic_candidate));
+    if (magic_candidate == TITAN_IMAGE_DESCRIPTOR_MAGIC) {
+      struct image_descriptor* img_dsc =
+          (struct image_descriptor*)(image + off);
+
+      if(img_dsc->descriptor_area_size + off > len) {
+        // Image descriptor is clipped
+        return NULL;
+      }
+
+      return img_dsc;
+    }
+  }
+  return NULL;
+}
+
+bool libhoth_payload_info(const uint8_t* image, size_t len,
+                          struct payload_info* payload_info) {
+  const struct image_descriptor* descr = libhoth_find_image_descriptor(image, len);
+  if (descr == NULL) {
+    return false;
+  }
+
+  memcpy(payload_info->image_name, descr->image_name,
+         sizeof(payload_info->image_name));
+  payload_info->image_name[sizeof(payload_info->image_name)-1] = 0;
+
+  payload_info->image_family = descr->image_family;
+  payload_info->image_version.major = descr->image_major;
+  payload_info->image_version.minor = descr->image_minor;
+  payload_info->image_version.point = descr->image_point;
+  payload_info->image_version.subpoint = descr->image_subpoint;
+  payload_info->image_type = descr->image_type;
+
+  if (descr->hash_type != HASH_SHA2_256) {
+    memset(payload_info->image_hash, 0, sizeof(payload_info->image_hash));
+  } else {
+    uint32_t region_size = descr->region_count * sizeof(struct image_region);
+    struct hash_sha256* hash =
+        (struct hash_sha256*)((uint8_t*)&descr->image_regions + region_size);
+    memcpy(payload_info->image_hash, hash->hash, sizeof(hash->hash));
+  }
+
+  return true;
+}
--- a/protocol/payload_info.h
+++ b/protocol/payload_info.h
@@ -1,0 +1,168 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBHOTH_PROTOCOL_PAYLOAD_INFO_H_
+#define LIBHOTH_PROTOCOL_PAYLOAD_INFO_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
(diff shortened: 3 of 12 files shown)


protocol/payload_info.c
```diff
@@ -0,0 +1,78 @@
...
+// limitations under the License.
+
+#include "payload_info.h"
+
+#include <string.h>
+
+const int64_t TITAN_IMAGE_DESCRIPTOR_MAGIC = 0x5F435344474D495F;
+const int64_t TITAN_IMAGE_DESCRIPTOR_ALIGNMENT = 1 << 16;
+
+const int64_t IMAGE_NAME_OFFSET = 20;
```
cjevans-google, line 22: Rather than including offsets, let's just import a subset of the `image_descriptor.h` structs -- `struct image_descriptor` and `struct image_region` (and a few of the associated enums).  We likely don't need the extension fields.

protocol/payload_info.c
```diff
@@ -0,0 +1,78 @@
...
+  payload_info->image_version.point = version_start[2];
+  payload_info->image_version.subpoint = version_start[3];
+
+  payload_info->image_type = descr_start[IMAGE_TYPE_OFFSET];
+
+  if (descr_start[HASH_TYPE_OFFSET] == 0) {
+    memset(payload_info->image_hash, 0, sizeof(payload_info->image_hash));
+  } else {
+    uint8_t* hash_start =
+        descr_start + 96                          /* region offset */
```
cjevans-google, line 69: These magic numbers will disappear with using `image_descriptor.h`.

cjevans-google (changes requested):

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
   }
   return 0;
 }
+
+int htool_payload_info(const struct htool_invocation* inv) {
+  struct libhoth_device* dev = htool_libhoth_device();
+  if (!dev) {
```
choumarin, line 82: Same as below
```suggestion
  if (dev == NULL) {
```

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+
+  uint8_t* image = mmap(NULL, statbuf.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+  if (image == MAP_FAILED) {
+    fprintf(stderr, "mmap error: %s\n", strerror(errno));
+    goto cleanup;
+  }
+
+  struct payload_info info;
+  bool res = libhoth_payload_info(image, statbuf.st_size, &info);
+  if (!res) {
```
choumarin, line 114: Since you're not using res, you can simplify to 
```suggestion
  if (!libhoth_payload_info(image, statbuf.st_size, &info)) {
```

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+         info.image_version.minor, info.image_version.point,
+         info.image_version.subpoint);
+  printf("  type: %u\n", info.image_type);
+  printf("  hash: ");
+  for (int i = 0; i < sizeof(info.image_hash); i++) {
+    printf("%x", info.image_hash[i]);
+  }
+  printf("\n");
+
+  int ret = munmap(image, statbuf.st_size);
```
xorptr, line 133: Would it be a good idea to move `munmap` to clean up as well? In a case where `libhoth_payload_info` fails, the mmap'd memory will not be unmapped gracefully

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+  printf("Payload Info:\n");
+  printf("  name: %-32s\n", info.image_name);
+  printf("  family: %u\n", info.image_family);
+  printf("  version: %u.%u.%u.%u\n", info.image_version.major,
+         info.image_version.minor, info.image_version.point,
+         info.image_version.subpoint);
+  printf("  type: %u\n", info.image_type);
+  printf("  hash: ");
+  for (int i = 0; i < sizeof(info.image_hash); i++) {
+    printf("%x", info.image_hash[i]);
```
xorptr, line 129: Would it be a good idea to add zeros on the left to pad to 2 characters?

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+
+#include <string.h>
+
+#define TITAN_IMAGE_DESCRIPTOR_MAGIC 0x5f435344474d495f  // "_IMGDSC_"
+#define HASH_MAGIC 0x48534148                            // "HASH"
+#define TITAN_IMAGE_DESCRIPTOR_ALIGNMENT (1 << 16)
+
+struct image_descriptor* libhoth_find_image_descriptor(uint8_t* image,
+                                                       size_t len) {
+  for (size_t off = 0; off + sizeof(int64_t) - 1 < len;
```
choumarin, line 25: Can we use sizeof(TITAN_IMAGE_DESCRIPTOR_MAGIC) so that it's not tied to the type?

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "payload_info.h"
+
+#include <string.h>
+
+#define TITAN_IMAGE_DESCRIPTOR_MAGIC 0x5f435344474d495f  // "_IMGDSC_"
+#define HASH_MAGIC 0x48534148                            // "HASH"
+#define TITAN_IMAGE_DESCRIPTOR_ALIGNMENT (1 << 16)
```
choumarin, line 21: Should we use types instead of #defines? eg
`static const uint64_t kTitanImageDescriptorMagic = 0x5f435344474d495f;`

At least we probably want to use UL ULL etc.

protocol/payload_info.c
choumarin: Should the loop end check also take into account the need for the image descriptor to fit? ie `+ sizeof(struct image_descriptor)`

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+    memcpy(&magic_candidate, image + off, sizeof(int64_t));
+    if (magic_candidate == TITAN_IMAGE_DESCRIPTOR_MAGIC) {
+      return (struct image_descriptor*)(image + off);
+    }
+  }
+  return NULL;
+}
+
+bool libhoth_payload_info(uint8_t* image, size_t len,
+                          struct payload_info* payload_info) {
```
choumarin, line 37: depending on who we're trusting to call this, should we check payload_info is nullptr?

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+
+#include "payload_info.h"
+
+#include <string.h>
+
+#define TITAN_IMAGE_DESCRIPTOR_MAGIC 0x5f435344474d495f  // "_IMGDSC_"
+#define HASH_MAGIC 0x48534148                            // "HASH"
+#define TITAN_IMAGE_DESCRIPTOR_ALIGNMENT (1 << 16)
+
+struct image_descriptor* libhoth_find_image_descriptor(uint8_t* image,
```
choumarin, line 23: depending on who we're trusting to call this, should we check image is nullptr?

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+  payload_info->image_version.point = descr->image_point;
+  payload_info->image_version.subpoint = descr->image_subpoint;
+  payload_info->image_type = descr->image_type;
+
+  if (descr->hash_type != HASH_SHA2_256) {
+    memset(payload_info->image_hash, 0, sizeof(payload_info->image_hash));
+  } else {
+    uint32_t region_size = descr->region_count * sizeof(struct image_region);
+    struct hash_sha256* hash =
+        (struct hash_sha256*)((uint8_t*)&descr->image_regions + region_size);
```
choumarin, line 58: if I understand correctly, we're copying from _after_ the image_descriptor, should we check against len to see if we have enough to read from?

protocol/payload_info.h
```diff
@@ -0,0 +1,160 @@
...
+  uint16_t region_version;
+  /* See IMAGE_REGION_* defines above. */
+  uint16_t region_attributes;
+} __attribute__((__packed__));
+
+/* Hash the static regions (IMAGE_REGION_STATIC) excluding this descriptor
+ * structure i.e. skipping image_descriptor.descriptor_size bytes (optional).
+ */
+struct hash_sha256 {
+  uint32_t hash_magic;  // #define HASH_MAGIC
```
choumarin, line 80:
```suggestion
  uint32_t hash_magic;
```

protocol/payload_info.h
```diff
@@ -0,0 +1,160 @@
...
+  uint8_t hash[32];
+} __attribute__((__packed__));
+
+struct hash_sha512 {
+  uint32_t hash_magic;  // #define HASH_MAGIC
```
choumarin, line 85:
```suggestion
  uint32_t hash_magic;
```

protocol/payload_info_test.cc
```diff
@@ -0,0 +1,66 @@
...
+  EXPECT_EQ(info.image_version.point, 0);
+  EXPECT_EQ(info.image_version.subpoint, 0);
+
+  EXPECT_EQ(info.image_type, 0);
+
+  std::stringstream stream;
+  stream << std::hex;
+  for (const auto c : info.image_hash) {
+    stream << std::setw(2) << std::setfill('0') << (int)c;
+  }
```
choumarin, line 61: if we have access to absl, you can use absl::BytesToHexString

protocol/payload_info.h
```diff
@@ -0,0 +1,160 @@
...
+  /* See IMAGE_REGION_* defines above. */
+  uint16_t region_attributes;
+} __attribute__((__packed__));
+
+/* Hash the static regions (IMAGE_REGION_STATIC) excluding this descriptor
+ * structure i.e. skipping image_descriptor.descriptor_size bytes (optional).
+ */
+struct hash_sha256 {
+  uint32_t hash_magic;  // #define HASH_MAGIC
+  uint8_t hash[32];
```
choumarin, line 81: maybe extract all the 32 as HASH_SHA256_BYTES?

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+    fprintf(stderr, "fstat error: %s\n", strerror(errno));
+    goto cleanup;
+  }
+  if (statbuf.st_size > SIZE_MAX) {
+    fprintf(stderr, "file too large\n");
+    goto cleanup;
+  }
+
+  uint8_t* image = mmap(NULL, statbuf.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+  if (image == MAP_FAILED) {
```
xorptr, line 107: Would it be a good idea to define this as `const uint8_t *` since the backing memory is read only?

protocol/payload_info.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+bool libhoth_payload_info(uint8_t* image, size_t len,
+                          struct payload_info* payload_info) {
+  struct image_descriptor* descr = libhoth_find_image_descriptor(image, len);
+  if (descr == NULL) {
+    return false;
+  }
+
+  memcpy(payload_info->image_name, descr->image_name,
```
xorptr, line 43: Would it be a good idea to add a NULL terminator at `payload_info->image_name[sizeof(payload_info->image_name)  -1]` just in case the file is malformed and there is no NULL terminator in `descr->image_name`?

protocol/payload_info.c
stevenportley: Done

examples/htool_payload.c
stevenportley: I don't want to munmap on an image that failed to originally map.  Cleaned this up to do the right thing, similar to some of the other htool command impls.

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+  if (ret != 0) {
+    fprintf(stderr, "munmap error: %d\n", ret);
+  }
+
+cleanup:
+  ret = close(fd);
+  if (ret != 0) {
+    fprintf(stderr, "close error: %d\n", ret);
+  }
+  return -1;
```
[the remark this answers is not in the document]
stevenportley: Done

protocol/payload_info.c
stevenportley: The sizeof(int64_t) is unnecessary if we are checking that it fits within the `image_descriptor`.  Fixed and added a more general size check.

protocol/payload_info.c
stevenportley: The additional check in the image descriptor search should cover this now.

examples/htool_payload.c
stevenportley: I'm not sure what you mean by this? This output is as follows, which I think is reasonable:
`Payload Info:
  name: test layout
  family: 2
  version: 1.0.0.0
  type: 0
  hash: 50316da1d3b06ff87989aa8bcd48a33ecc4bfe2114ded138ee594f697d58b3
`

protocol/payload_info_test.cc
stevenportley: I don't have absl and I would hesitate to pull it in just for this.

protocol/payload_info.h
stevenportley: These comments came from the original file and I think they're fairly helpful so I would like to leave them in.  Updated them to match the new name

protocol/payload_info.h
stevenportley: Same as above

protocol/payload_info.h
stevenportley: Done.

examples/htool_payload.c
stevenportley: I cannot make this const because munmap expects a non-const pointer.  Regardless, I've const-ified the rest of the API since it makes sense to do so.

protocol/payload_info.c
stevenportley: Yeah, I'm returning NULL if an image descriptor is not found.

protocol/payload_info.c
choumarin: If I understand correctly, the hashes come _after_ the image descriptor, so you need to check len for that.
Also, the image descriptor's `image_regions[]` is unsized, so it's not getting counted in `sizeof(struct image_descriptor)`.

choumarin: ahh never mind, that's what `descriptor_area_size` is for I think

examples/htool_payload.c
stevenportley: Ah I see, good catch. Fixed.

examples/htool_payload.c
```diff
@@ -68,3 +76,69 @@ int htool_payload_status() {
...
+  printf("  type: %u\n", info.image_type);
+  printf("  hash: ");
+  for (int i = 0; i < sizeof(info.image_hash); i++) {
+    printf("%x", info.image_hash[i]);
+  }
+  printf("\n");
+
+  int ret = munmap(image, statbuf.st_size);
+  if (ret != 0) {
+    fprintf(stderr, "munmap error: %d\n", ret);
```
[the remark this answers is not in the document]
stevenportley: Yes, `ret` isn't useful here.  Fixed.

cjevans-google (approved):

PR #93, opened 2025-03-19
Add support for finding libhoth USB devices

stevenportley:


transports/libhoth_usb.h
```diff
@@ -47,6 +55,12 @@ int libhoth_usb_receive_response(struct libhoth_device* dev, void* response,
                                  size_t max_response_size, size_t* actual_size,
                                  int timeout_ms);
 
+bool libhoth_device_is_hoth(struct libusb_device_descriptor* dev);
+int libhoth_usb_get_device(libusb_context* ctx,
```
xorptr, line 59: Where is this function used?

stevenportley: It's currently not used right now.  htool has a similar implementation in htool_usb that has additional complexity that I was trying to avoid bringing into the public libhoth API.  If you think we might want `enumerate_devices` and the filters for other libhoth clients, I could add it.

PR #94, opened 2025-03-20
Return code 0 when payload update protocol succeeds (regardless of cleanup errors)

xorptr:
Folks using the tool noticed that exit code for htool changed from 0 to 255 when payload update succeeds. Fixing the tool to exit with code 0 in case payload update succeeds.

### DIFF
--- a/examples/htool_payload_update.c
+++ b/examples/htool_payload_update.c
@@ -18,14 +18,12 @@
 #include <fcntl.h>
 #include <stdint.h>
 #include <stdio.h>
-#include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


stevenportley (approved):
I think this is good regardless, but what cleanup errors are you seeing? munmap or close?

xorptr:
> I think this is good regardless, but what cleanup errors are you seeing? munmap or close?

I don't think there were cleanup errors. My understanding is that the flow I changed used to return -1 even if cleanup was successfully executed.

xorptr:
I can change the logic to return cleanup error code (if any) if we want to do so

stevenportley:
I think success is preferred if payload update passes

PR #96, opened 2025-03-24
Add controlled storage to protocol/

stevenportley:


examples/htool.c
```diff
@@ -1047,6 +1157,34 @@ static const struct htool_cmd CMDS[] = {
                 {}},
         .func = command_jtag_operation_run,
     },
+    {
+        .verbs = (const char*[]){"storage", "read", NULL},
+        .desc = "Read from the controlled storage",
+        .params =
+            (const struct htool_param[]){
+                {HTOOL_FLAG_VALUE, 's', "slot", "0", .desc = "slot"},
```
choumarin, line 1165:
```suggestion
                {HTOOL_FLAG_VALUE, .ch = 's', .name="slot", .default="0", .desc = "slot"},
```

protocol/controlled_storage.c
```diff
@@ -0,0 +1,57 @@
...
+  req.slot = slot;
+  return libhoth_hostcmd_exec(
+      dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HAVEN_CONT_STORAGE,
+      /*version=*/0, &req, sizeof(req), payload, sizeof(*payload), payload_len);
+}
+
+int libhoth_controlled_storage_write(struct libhoth_device* dev, uint32_t slot,
+                                     const uint8_t* data, size_t len) {
+  struct ec_request_controlled_storage req = {};
+  if (len > CONT_STORAGE_SIZE) {
```
choumarin, line 34: nit technically the same but I think this is more robust
```suggestion
  if (len > sizeof(req.payload.data)) {
```

protocol/controlled_storage.h
```diff
@@ -0,0 +1,62 @@
...
+
+// Serves as the response to CONT_STORAGE_READ as well as payload to CONT_STORAGE_WRITE.
+struct ec_payload_controlled_storage {
+  uint8_t data[CONT_STORAGE_SIZE];
+} __attribute__((packed));
+
+struct ec_request_controlled_storage {
+  /* The operation is one of controlled_storage_op. */
+  uint32_t operation;
+  /* Choose which (0-5) slot to operate on. */
```
choumarin, line 45: is the (0-5) limit implemented in the fw? I don't see restrictions in this PR

choumarin (approved):

protocol/controlled_storage.h
stevenportley: Yes, this is a firmware imposed limit.

examples/htool.c
```diff
@@ -622,6 +624,115 @@ int htool_external_usb_host_check_presence(const struct htool_invocation* inv) {
...
+  if (file_data == NULL) {
+    goto cleanup;
+  }
+
+  int ret = libhoth_controlled_storage_write(dev, slot, file_data, file_size);
+  munmap(file_data, file_size);
+
+cleanup:
+  close(fd);
+  return ret;
```
xorptr, line 669: I see that `ret` is declared and defined at line 664, but `cleanup` label can be reached from line 651. Would this cause any issues?

protocol/controlled_storage.c
```diff
@@ -0,0 +1,57 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "protocol/controlled_storage.h"
+
+#include <string.h>
+
+int libhoth_controlled_storage_read(
+    struct libhoth_device* dev, uint32_t slot,
+    struct ec_payload_controlled_storage* payload, size_t* payload_len) {
```
xorptr, line 21: Would it be a good idea to add NULL checks on parameters for public APIs?

protocol/controlled_storage.c
```diff
@@ -0,0 +1,57 @@
...
+
+int libhoth_controlled_storage_delete(struct libhoth_device* dev,
+                                      uint32_t slot) {
+  struct ec_request_controlled_storage req = {};
+
+  req.operation = CONTROLLED_STORAGE_DELETE;
+  req.slot = slot;
+  return libhoth_hostcmd_exec(
+      dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HOTH_CONTROLLED_STORAGE,
+      /*version=*/0, &req, sizeof(req), NULL, 0, NULL);
```
xorptr, line 56: Does the tool need to send the `req.payload.data` in this case?

protocol/controlled_storage.c
```diff
@@ -0,0 +1,57 @@
...
+int libhoth_controlled_storage_read(
+    struct libhoth_device* dev, uint32_t slot,
+    struct ec_payload_controlled_storage* payload, size_t* payload_len) {
+  struct ec_request_controlled_storage req = {};
+
+  req.operation = CONTROLLED_STORAGE_READ;
+  req.slot = slot;
+  return libhoth_hostcmd_exec(
+      dev, EC_CMD_BOARD_SPECIFIC_BASE + EC_PRV_CMD_HOTH_CONTROLLED_STORAGE,
+      /*version=*/0, &req, sizeof(req), payload, sizeof(*payload), payload_len);
```
xorptr, line 28: Does the tool need to send `req.payload.data` in for read operation?

protocol/controlled_storage.c
stevenportley: We haven't done this for any of the other protocols so I would like to keep it consistent.

protocol/controlled_storage.c
stevenportley: It apparently does.  The read command is rejected if the payload is empty.

protocol/controlled_storage.c
xorptr: Ack

examples/htool.c
stevenportley: Good catch, fixed.

protocol/controlled_storage.c
xorptr: Ack. This requirement should probably be investigated for removal later.

protocol/controlled_storage_test.cc
```diff
@@ -0,0 +1,91 @@
...
+      .WillOnce(DoAll(CopyResp(&ex_resp, ex_payload_len), Return(LIBHOTH_OK)));
+
+  struct hoth_payload_controlled_storage resp = {};
+  size_t payload_len = 0;
+
+  EXPECT_EQ(libhoth_controlled_storage_read(&hoth_dev_, 0, &resp, &payload_len),
+            LIBHOTH_OK);
+
+  EXPECT_EQ(payload_len, ex_payload_len);
+  EXPECT_EQ(memcmp(&ex_resp, &resp, payload_len), 0);
```
xorptr, line 52: Would it be a good idea to use `testing::ElementsAreArray` here? I think that prints the elements if the test fails which `memcmp` won't do.

xorptr (approved):

PR #100, opened 2025-04-04
Refactor JTAG operations to use new code structure

xorptr:


protocol/jtag.c
```diff
@@ -0,0 +1,168 @@
...
+
+int libhoth_jtag_program_and_verify_pld(struct libhoth_device* dev,
+                                        uint32_t offset) {
+  struct {
+    struct hoth_request_jtag_operation operation;
+    struct hoth_request_jtag_program_and_verify_pld_operation params;
+  } __attribute__((packed, aligned(4))) request = {
+      .operation =
+          {
+              .clk_idiv = 0,  // Not used
```
xorptr, line 103: This is not set since it is unused for now (same behavior as before re-structuring). But I might add it in later pull request

protocol/jtag.c
```diff
@@ -0,0 +1,168 @@
...
+}
+
+int libhoth_jtag_verify_pld(struct libhoth_device* dev, uint32_t offset) {
+  struct {
+    struct hoth_request_jtag_operation operation;
+    struct hoth_request_jtag_program_and_verify_pld_operation params;
+  } __attribute__((packed, aligned(4))) request = {
+      .operation =
+          {
+              .clk_idiv = (uint16_t)0,  // Not used
```
xorptr, line 140: This is not set since it is unused for now (same behavior as before re-structuring). But I might add it in later pull request

cjevans-google (approved):

PR #105, opened 2025-05-13
Fix MSAN failures

stevenportley:
`unused` here is still used within the mock, causing msan failures.


protocol/jtag_test.cc
```diff
@@ -222,7 +222,7 @@ TEST_F(LibHothTest, jtag_program_and_verify_pld_receive_unexpected_length) {
                           _))
       .WillOnce(Return(LIBHOTH_OK));
 
-  uint8_t unused;
```
xorptr, line 225: Would it be a good idea to rename it to indicate that it is used? Something like `fake_response_data`?

stevenportley: Yeah, probably a good idea.

PR #113, opened 2025-05-30
format the meson.build

zhdaniel12:
format the meson.build with `meson format -i -r`


zhdaniel12:
will update with a new pull request which format meson and install header to libhoth

PR #211, opened 2026-02-18
Update readme with dependencies and updated build instructions

mfield4:
- Update readme with dependencies and updated build instructions.
- Update htool output to reflect current state of the tool.


README.md
xorptr: Which version of `htool` did you use here? I see missing fields and options on the following commands

xorptr (approved):

PR #219, opened 2026-02-26
[protocol] Add the payload update confirm features cmds

esnguyen:
This seeks to add payload update confirm cmds for get, enable, disable,
and confirm. This are the standard feature that is needed to interact
with payload update confirm.

### DIFF
--- a/examples/htool.c
+++ b/examples/htool.c
@@ -997,6 +997,31 @@ static const struct htool_cmd CMDS[] = {
         .func = htool_payload_update,
     },
     {
+        .verbs = (const char*[]){"payload", "confirm", NULL},
+        .desc = "Finish a payload update confirmation.",
+        .params = (const struct htool_param[]){{}},
+        .func = htool_payload_update_confirm,
+    },
+    {
+        .verbs = (const char*[]){"payload", "get_timeout", NULL},
+        .desc = "Get the current payload update confirmation timeout.",
+        .params = (const struct htool_param[]){{}},
+        .func = htool_payload_update_confirm_get_staged_timeout,
+    },
+    {
+        .verbs = (const char*[]){"payload", "timeout_enable", NULL},
+        .desc =
+            "Enable/Disable the timeout mechanism for payload update confirm",
+        .params =
+            (const struct htool_param[]){
+                {HTOOL_FLAG_BOOL, 'e', "enable", "true",
+                 .desc = "Enable confirm mode"},
+                {HTOOL_FLAG_VALUE, 't', "timeout", "0",
+                 .desc = "Timeout in seconds"},
+                {}},
+        .func = htool_payload_update_confirm_enable,
+    },
+    {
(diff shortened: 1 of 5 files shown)


protocol/payload_update.h
```diff
@@ -39,6 +40,50 @@ extern "C" {
 #define PAYLOAD_UPDATE_CONFIRM 10
 #define PAYLOAD_UPDATE_VERIFY_DESCRIPTOR 11
```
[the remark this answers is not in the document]
esnguyen: @korran sure this makes sense. I relocate them within the payload_update.c since most of them are there.

examples/htool_payload_update.c
```diff
@@ -73,6 +83,15 @@ int htool_payload_update(const struct htool_invocation* inv) {
     goto cleanup;
   }
 
+  if (timeout > PAYLOAD_UPDATE_CONFIRM_SECONDS_MAX ||
```
rkr35: While checking the seconds here provides better UX error messages, we would have to keep these timeouts in sync between `htool` and the different RoT firmwares (Dauntless, EarlGrey).

Consider removing this check and letting the firmware return the appropriate error codes.

rkr35: Another concern with syncing the timeouts is that we may relax the timeouts in the firmware but forget to update the same timeouts in `htool`, which creates a UX sharp edge where `htool` is being unnecessarily strict.

examples/htool.c
```diff
@@ -985,17 +985,34 @@ static const struct htool_cmd CMDS[] = {
...
+        .desc = "Perform payload update protocol for Titan images. Optional "
+                "allows for payload confirmation.",
         .params =
             (const struct htool_param[]){
                 {HTOOL_POSITIONAL, .name = "source-file"},
                 {HTOOL_FLAG_BOOL, 's', "skip_erase", "false",
                  .desc = "Skip erasing the staging side."},
                 {HTOOL_FLAG_BOOL, 'b', "binary", "false",
                  .desc = "Update with generic binary file"},
+                {HTOOL_FLAG_BOOL, 'e', "enable", "false",
```
rkr35: To keep the `payload update` command smaller, consider splitting the "enable confirm" functionality to a separate command.

examples/htool_payload_update.c
```diff
@@ -286,3 +314,40 @@ int htool_payload_update_verify(const struct htool_invocation* inv) {
...
+
+  payload_update_confirm_response_t response = {0};
+
+  int ret = libhoth_payload_update_confirm_get_timeout(dev, &response);
+  if (ret != 0) {
+    fprintf(stderr, "Failed to get payload update timeout\n");
+    return -1;
+  }
+
+  printf("Current timeout: %u seconds\n", response.timeouts.current);
```
rkr35, line 322: Consider printing out all the values in `response.timeouts`.

rkr35 (approved):

PR #243, opened 2026-04-03
Add host command to enable/disable JTAG debug

aranikam:
Note: This will stop working once authorized host commands are supported

### DIFF
--- a/examples/htool.c
+++ b/examples/htool.c
@@ -32,6 +32,7 @@
 #include "htool_authz_command.h"
 #include "htool_cmd.h"
 #include "htool_console.h"
+#include "htool_debug.h"
 #include "htool_dfu.h"
 #include "htool_firmware_update.h"
 #include "htool_i2c.h"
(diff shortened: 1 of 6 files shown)


examples/htool_debug.c
```diff
@@ -0,0 +1,64 @@
+
+// Copyright 2025 Google LLC
```
xorptr: nit: 2026

examples/htool_debug.h
```diff
@@ -0,0 +1,31 @@
+// Copyright 2025 Google LLC
```
xorptr: nit: 2026

examples/htool.c
```diff
@@ -928,6 +929,24 @@ static const struct htool_cmd CMDS[] = {
         .params = (const struct htool_param[]){{}},
         .func = command_target_reset_pulse,
     },
+    {
+        .verbs = (const char*[]){"target_debug", "enable", NULL},
```
xorptr: nit: Is this intentionally different from `target reset` command structure? Would it be a good idea to use the same structure (i.e. `target reset` and `target debug`)?

xorptr (approved):

PR #245, opened 2026-04-15
[protocol] Add header and api for mauv

esnguyen:
We are planning to have mauv support to get the compiled firmware and payload in htool. The main point is to align the functionality that haventool has for mauv. Htool should have this as well.

### DIFF
--- a/protocol/mauv.c
+++ b/protocol/mauv.c
@@ -1,0 +1,22 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "mauv.h"
+
+int libhoth_get_mauv(struct libhoth_device* dev, uint8_t state,
+                     uint8_t category, struct hoth_response_mauv* mauv) {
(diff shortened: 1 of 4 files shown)


protocol/mauv.h
```diff
@@ -0,0 +1,197 @@
...
+              "mauv_version should be at offset 4");
+static_assert(offsetof(struct firmware_mauv,
+                       minimum_acceptable_update_version) == 8,
+              "minimum_acceptable_update_version should be at offset 8");
+static_assert(offsetof(struct firmware_mauv, denylist_num_entries) == 24,
+              "denylist_num_entries should be at offset 24");
+static_assert(offsetof(struct firmware_mauv, denylist) == 28,
+              "denylist should be at offset 28");
+
+struct payload_mauv {
```
pbadre: I recommend have the struct names same as what we have in firmware. This is image_mauv currently (Current naming is very ambiguous and was a wrong choice but I prefer to keep it uniform on both sides and change everything later?) 
Also recommend to add Lint if change in firmware header file to change this file if struct changes

protocol/mauv.c
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "mauv.h"
+
+int libhoth_get_mauv(struct libhoth_device* dev, uint8_t state,
```
pbadre, line 17: I recommend having the same format as haventool (Also applies for set_mauv in the future). Currently this function is called fetch_mauv and present in fetch_mauv file in haventool. Perhaps we want to have separate files for get vs set (Each of these have 4 subcommands each)

esnguyen: @pbadre sounds good! Ill get appended the libhoth_ with the format of haventool. I prefer to keep everything in set & get to be align with htool.

pbadre (approved):

PR #248, opened 2026-04-16
[htool] Add support to payload for payload mauv compile

esnguyen:
We are adding to the existing payload command to allow for an additional
option of reading back the compiled payload mauv. This is a set of
features that are aligned with the overall mauv support.


cjevans-google (approved):
nit: Commit message should say "compiled" not "compile" (verb)

PR #249, opened 2026-04-25
[transport] usb: Fix htool crash on USB disconnect

wak-google:
Previously, libhoth_usb_fifo_run_transfers() could return while one or more libusb transfers were still pending. This happened if the second transfer submission failed, or if the event handling loop was interrupted by a signal (e.g. SIGINT).

When htool subsequently attempted to close and reopen the transport, it would call libusb_free_transfer() on these pending transfers, triggering an assertion failure in libusb:
'usbi_mutex_lock: Assertion pthread_mutex_lock(mutex) == 0 failed'

This change ensures that:
1. All submitted transfers are completed (success, error, or cancel) before the function returns.
2. If a transfer submission fails, any other successfully submitted transfer is cancelled and waited for.
3. Signal interruptions do not cause an early return while transfers are pending.
4. Completion flags are correctly initialized on open.

Verified on yutulis-ru4-bmc-01 with 1000 iterations of target reset spam without a crash.

### DIFF
--- a/transports/libhoth_usb_fifo.c
+++ b/transports/libhoth_usb_fifo.c
@@ -37,20 +37,40 @@ static int libhoth_usb_fifo_run_transfers(struct libhoth_usb_device* dev,
   if (in) {
     int status = libusb_submit_transfer(drvdata->in_transfer);
     if (status != LIBUSB_SUCCESS) {
+      drvdata->all_transfers_completed = 1;
+      drvdata->in_transfer_completed = true;
+      drvdata->out_transfer_completed = true;
       return status;
     }
   }
   if (out) {
     int status = libusb_submit_transfer(drvdata->out_transfer);
     if (status != LIBUSB_SUCCESS) {
+      if (in) {
+        libusb_cancel_transfer(drvdata->in_transfer);
+        // We must still wait for the IN transfer to complete/cancel
+        while (drvdata->all_transfers_completed == 0) {
+          libusb_handle_events_completed(dev->ctx,
+                                         &drvdata->all_transfers_completed);
+        }
+      } else {
+        drvdata->all_transfers_completed = 1;
+        drvdata->out_transfer_completed = true;
+      }
       return status;
     }
   }
   while (drvdata->all_transfers_completed == 0) {
     int status = libusb_handle_events_completed(
         dev->ctx, &drvdata->all_transfers_completed);
-    if (status == LIBUSB_ERROR_INTERRUPTED) {
-      return status;
+    if (status != LIBUSB_SUCCESS && status != LIBUSB_ERROR_INTERRUPTED) {
+      // On a real error, try to cancel everything to speed up completion
+      if (!drvdata->in_transfer_completed) {
+        libusb_cancel_transfer(drvdata->in_transfer);
+      }
+      if (!drvdata->out_transfer_completed) {
+        libusb_cancel_transfer(drvdata->out_transfer);
+      }
     }
   }
   return LIBHOTH_OK;
@@ -159,6 +179,9 @@ int libhoth_usb_fifo_open(struct libhoth_usb_device* dev,
     goto err_out;
   }
   drvdata->prng_state = prng_seed;
+  drvdata->in_transfer_completed = true;
+  drvdata->out_transfer_completed = true;
+  drvdata->all_transfers_completed = 1;
   return LIBHOTH_OK;
 err_out:
   if (drvdata->in_buffer != NULL) free(drvdata->in_buffer);
--- a/transports/libhoth_usb_fifo.c
+++ b/transports/libhoth_usb_fifo.c
@@ -29,31 +29,50 @@
 
 static int libhoth_usb_fifo_run_transfers(struct libhoth_usb_device* dev,
                                           bool out, bool in) {
+  int status;
   struct libhoth_usb_fifo* drvdata = &dev->driver_data.fifo;
-  drvdata->all_transfers_completed = 0;
-  drvdata->out_transfer_completed = !out;
-  drvdata->in_transfer_completed = !in;
+  drvdata->all_transfers_completed = 1;
+  drvdata->out_transfer_completed = true;
+  drvdata->in_transfer_completed = true;
 
   if (in) {
-    int status = libusb_submit_transfer(drvdata->in_transfer);
+    status = libusb_submit_transfer(drvdata->in_transfer);
     if (status != LIBUSB_SUCCESS) {
-      return status;
+      goto out;
     }
+    drvdata->all_transfers_completed = 0;
+    drvdata->in_transfer_completed = false;
   }
   if (out) {
-    int status = libusb_submit_transfer(drvdata->out_transfer);
+    status = libusb_submit_transfer(drvdata->out_transfer);
     if (status != LIBUSB_SUCCESS) {
-      return status;
+      goto out;
     }
+    drvdata->all_transfers_completed = 0;
+    drvdata->out_transfer_completed = false;
   }
   while (drvdata->all_transfers_completed == 0) {
-    int status = libusb_handle_events_completed(
-        dev->ctx, &drvdata->all_transfers_completed);
-    if (status == LIBUSB_ERROR_INTERRUPTED) {
-      return status;
+    status = libusb_handle_events_completed(dev->ctx,
+                                            &drvdata->all_transfers_completed);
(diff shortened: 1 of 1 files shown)


transports/libhoth_usb_fifo.c
```diff
@@ -159,6 +179,9 @@ int libhoth_usb_fifo_open(struct libhoth_usb_device* dev,
     goto err_out;
   }
   drvdata->prng_state = prng_seed;
+  drvdata->in_transfer_completed = true;
```
xorptr, line 183: nit: Should these be initialized as `false`/`0`?

transports/libhoth_usb_fifo.c
```diff
@@ -37,20 +37,40 @@ static int libhoth_usb_fifo_run_transfers(struct libhoth_usb_device* dev,
   if (in) {
     int status = libusb_submit_transfer(drvdata->in_transfer);
     if (status != LIBUSB_SUCCESS) {
+      drvdata->all_transfers_completed = 1;
+      drvdata->in_transfer_completed = true;
+      drvdata->out_transfer_completed = true;
```
xorptr: nit: is there a reason to modify this here?

transports/libhoth_usb_fifo.c
```diff
@@ -37,20 +37,40 @@ static int libhoth_usb_fifo_run_transfers(struct libhoth_usb_device* dev,
...
       return status;
     }
   }
   if (out) {
     int status = libusb_submit_transfer(drvdata->out_transfer);
     if (status != LIBUSB_SUCCESS) {
+      if (in) {
+        libusb_cancel_transfer(drvdata->in_transfer);
+        // We must still wait for the IN transfer to complete/cancel
+        while (drvdata->all_transfers_completed == 0) {
```
xorptr: Should `drvdata->out_transfer_complete` be set to `true` before the loop? The code in `fifo_transfer_callback` sets `all_transfers_completed` to `1` only if both `in_transfer_completed` and `out_transfer_completed` are true. `in_transfer_completed` will be set to `true` by `fifo_transfer_callback` if the submission was successful. I am not sure what else would set `out_transfer_completed` to `true`

```c
  if (drvdata->in_transfer_completed && drvdata->out_transfer_completed) {
    drvdata->all_transfers_completed = 1;
  }
```

transports/libhoth_usb_fifo.c
```diff
@@ -37,20 +37,40 @@ static int libhoth_usb_fifo_run_transfers(struct libhoth_usb_device* dev,
...
     }
   }
   while (drvdata->all_transfers_completed == 0) {
     int status = libusb_handle_events_completed(
         dev->ctx, &drvdata->all_transfers_completed);
-    if (status == LIBUSB_ERROR_INTERRUPTED) {
-      return status;
+    if (status != LIBUSB_SUCCESS && status != LIBUSB_ERROR_INTERRUPTED) {
+      // On a real error, try to cancel everything to speed up completion
+      if (!drvdata->in_transfer_completed) {
```
xorptr: Should we check if `in` and `out` were transfers were submitted before canceling respective transfers?

transports/libhoth_usb_fifo.c
wak-google: When we initialize, there is no transfer in progress so there is nothing to clean up

transports/libhoth_usb_fifo.c
wak-google: You don't need to cancel a tranfer if it failed, it's already canceled and you would be double freeing.

transports/libhoth_usb_fifo.c
wak-google: Good point, I think we always want to consider all transfers completed after we cancel and wait

transports/libhoth_usb_fifo.c
wak-google: The have to be submitted successfully to get to these points in the code

wak-google:
Should be fixed up now

transports/libhoth_usb_fifo.c
```diff
@@ -29,31 +29,52 @@
...
   }
-  return LIBHOTH_OK;
+  drvdata->in_transfer_completed = true;
+  drvdata->out_transfer_completed = true;
+  status = LIBHOTH_OK;
+
+out:
+  if (!drvdata->in_transfer_completed) {
+    libusb_cancel_transfer(drvdata->in_transfer);
+    drvdata->in_transfer_completed = true;
```
xorptr: Should this be handled in `fifo_transfer_callback` itself? I am assuming that `libusb_cancel_transfer` will cause callbacks to occur with status `LIBUSB_TRANSFER_CANCELLED`. Marking `in_transfer_completed` and `out_transfer_completed` here might cause `fifo_transfer_callback` to set `all_transfers_completed` before the both the transfers are actually cancelled.

In the previous implementation, this is my current understanding of the flow (assuming both transfers are submitted successfully, but one of them had an issue):
1. IN transfer has some error
2. `fifo_transfer_callback` is called with `transfer = drvdata->in_transfer` (assumed for this example), and `transfer->status = LIBUSB_TRANSFER_ERROR` (assumed for this example).
3. `fifo_transfer_callback` sets ` drvdata->in_transfer_completed = true`
4. `fifo_transfer_callback` calls `libusb_cancel_transfer(drvdata->out_transfer)`
5. `fifo_transfer_callback` returns without setting `all_transfers_completed` to true
6. Out transfer is cancellation processing is finished
7. `fifo_transfer_callback` is called with `transfer = drvdata->out_transfer` (assumed for this example), and `transfer->status = LIBUSB_TRANSFER_CANCELLED`
8. `fifo_transfer_callback` sets `drvdata->out_transfer_completed = true`
9. `fifo_transfer_callback` sets `drvdata->all_transfers_completed = 1`
10. Execution breaks out of loop with condition `drvdata->all_transfers_completed`

By setting both `drvdata->in_transfer_completed` and `drvdata->out_transfer_completed` just after submitting the transfer cancellation request, `fifo_transfer_callback` will set `drvdata->all_transfers_completed = 1` even if cancellation for just one of them is finished.

wak-google: Understood, check my latest update

transports/libhoth_usb_fifo.c
```diff
@@ -29,31 +29,50 @@
...
+
+out:
+  if (!drvdata->in_transfer_completed) {
+    libusb_cancel_transfer(drvdata->in_transfer);
+  }
+  if (!drvdata->out_transfer_completed) {
+    libusb_cancel_transfer(drvdata->out_transfer);
+  }
+  while (drvdata->all_transfers_completed == 0) {
+    libusb_handle_events_completed(dev->ctx, &drvdata->all_transfers_completed);
```
stevenportley, line 73: Should we look at the return type of this? I'm worried about getting stuck in an infinite loop here.

wak-google: It doesn't really matter if there is an error reported. Even if that happens we still need to continue polling until libusb is satisfied that all outstanding transfers are complete. We can never free the transfer data until the completion has been registered. If one fails you might even have another outstanding.

stevenportley (approved):

transports/libhoth_usb_fifo.c
```diff
@@ -29,31 +29,50 @@
...
+    status = libusb_submit_transfer(drvdata->in_transfer);
     if (status != LIBUSB_SUCCESS) {
-      return status;
+      goto out;
     }
+    drvdata->all_transfers_completed = 0;
+    drvdata->in_transfer_completed = false;
   }
   if (out) {
     int status = libusb_submit_transfer(drvdata->out_transfer);
```
xorptr: nit: this declares a new `status` variable inside this scope. Should this assignment occur to the `status` outside the scope (similar to what is done for `in` transfers)?

transports/libhoth_usb_fifo.c
```diff
@@ -29,31 +29,50 @@
...
     if (status != LIBUSB_SUCCESS) {
-      return status;
+      goto out;
     }
+    drvdata->all_transfers_completed = 0;
+    drvdata->out_transfer_completed = false;
   }
   while (drvdata->all_transfers_completed == 0) {
     int status = libusb_handle_events_completed(
```
xorptr: nit: Same concern as for `out` transfer.

xorptr (approved):

PR #251, opened 2026-04-28
Add the support for the new chipinfo format (OpenTitan) in htool.

paipeng-g:


paipeng-g:
Sample output on dev board: 

$ bazel-bin/examples/htool --usb_loc 5-1 show chipinfo
Chip Info:
OpenTitan iSerial: 0x0100000047425f54000000014742000000000000000340360500446400024001
OpenTitan Device ID: 0x000340360500446400024001

Generic Identifier:
  SI Creator ID:         0x4001
  Product ID:            0x0002
  Device Date:           2024 week 46
  Lot Number:            004
  Wafer Number:          05
  Wafer X Coord:         036
  Wafer Y Coord:         034
  Reserved DIN:          0x00
  Reserved:              0x00000000

SKU-specific Identifier:
  Package ID:            0
  AST Config Version:    0
  OTP ID:                "GB"
  OTP Version:           1
  SKU ID String:         "GB_T"
  SKU Specific Version:  1

Sample output on Asahi:

root@asahi:/tmp# ./htool --transport spidev --spidev_path /dev/spidev1.0 --mailbox_location 0x7ff0000 show chipinfo
Chip Info:
OpenTitan iSerial: 0x01000000544930300000000054490d0300000000000100350900811500034001
OpenTitan Device ID: 0x000100350900811500034001

Generic Identifier:
  SI Creator ID:         0x4001
  Product ID:            0x0003
  Device Date:           2025 week 11
  Lot Number:            008
  Wafer Number:          09
  Wafer X Coord:         035
  Wafer Y Coord:         010
  Reserved DIN:          0x00
  Reserved:              0x00000000

SKU-specific Identifier:
  Package ID:            3
  AST Config Version:    13
  OTP ID:                "TI"
  OTP Version:           0
  SKU ID String:         "TI00"
  SKU Specific Version:  1

stevenportley:
Can you merge these 6 commits into one? This way they don't all end up in the history

protocol/chipinfo.h
```diff
@@ -24,13 +24,22 @@ extern "C" {
 #include "transports/libhoth_device.h"
 
 #define HOTH_PRV_CMD_HOTH_CHIP_INFO 0x0010
-
-struct hoth_response_chip_info {
+struct libhoth_haven_device_id {
```
stevenportley: We've scrubbed this repo of this keyword and use `hoth` instead.  Can you rename this to something else?

examples/htool.c
```diff
@@ -112,16 +112,64 @@ static int command_show_chipinfo(const struct htool_invocation* inv) {
...
+    return -1;
+  }
+
+  uint8_t* id = response.data.open_titan_device_id;
+  printf("OpenTitan iSerial: 0x");
+  for (int i = 0; i < 32; i++) {
+    printf("%02x", id[i]);
+  }
+  printf("\nOpenTitan Device ID: 0x");
+  for (int i = 20; i < 32; i++) {
```
stevenportley, line 174: If you don't want to use a packed struct, can you at least use separate functions for interpreting the different parts of the opentitan serial? Even just a one line function that takes the serial as a parameter and returns the device id / wafer info / etc would make this more readable.

examples/htool.c
```diff
@@ -215,6 +263,11 @@ static int command_authz_host_command_build(
     fprintf(stderr, "Failed to get chip ID. status=%d\n", status);
     return -1;
   }
+  if (chipinfo_resp.version != 0) {
```
stevenportley, line 291: Will this break support for older fw versions?

stevenportley (changes requested):

examples/htool.c
paipeng-g: This is for authz_host_command and I haven't added the support for EG.

protocol/chipinfo.h
paipeng-g: Sure. Renamed to hoth

examples/htool.c
paipeng-g: Added a new structure.

protocol/chipinfo.h
```diff
@@ -24,13 +24,44 @@ extern "C" {
 #include "transports/libhoth_device.h"
 
 #define HOTH_PRV_CMD_HOTH_CHIP_INFO 0x0010
-
-struct hoth_response_chip_info {
+struct libhoth_hoth_device_id {
```
stevenportley: I think the `libhoth_hoth_` might be redundant?

protocol/chipinfo.c
```diff
@@ -14,11 +14,69 @@
...
+  int ret = libhoth_hostcmd_exec(
       dev, HOTH_CMD_BOARD_SPECIFIC_BASE + HOTH_PRV_CMD_HOTH_CHIP_INFO,
-      /*version=*/0, NULL, 0, chipinfo, sizeof(*chipinfo), NULL);
+      /*version=*/0, NULL, 0, resp_buf, sizeof(resp_buf), &resp_size);
+
+  if (ret != 0) {
+    return ret;
+  }
+
+  if (resp_size == 16) {
```
stevenportley: Can you defined constants for `16` and `32` here?

protocol/chipinfo.h
```diff
@@ -24,13 +24,44 @@ extern "C" {
...
+
+int parse_opentitan_device_id(const uint8_t* src,
+                              struct opentitan_device_id* dst);
+
+struct hoth_response_chip_info {
+  uint32_t version;  // 0: old format, 1: OpenTitan
+  uint32_t reserved;
+  union {
+    struct libhoth_hoth_device_id hoth_device_id;  // Old format (16 bytes)
+    uint8_t open_titan_device_id[32];  // New format (32 bytes) - OpenTitan
```
stevenportley: Instead of just a chunk of bytes, maybe we can use the actual struct here? 

If so, we would probably want to add a bunch of assertions for offset + size similar to how dfu_hostcmd.h and opentitan_version.h do it.  Would make it easier to work with rather than having a separate `parse_opentitan_device_id` function.

stevenportley (approved):

protocol/chipinfo.h
paipeng-g: Done.

protocol/chipinfo.h
paipeng-g: We cannot directly use the struct since some fields in the 32-byte device ID are not aligned to the byte boundary.

protocol/chipinfo.c
paipeng-g: Changed "32" to "OPENTITAN_DEVICE_ID_LEN", and "16" to "sizeof(struct hoth_device_id)".

PR #253, opened 2026-05-19
Update version checks during DFU update

xorptr:
- Do not fail desired ROM_EXT version checks if the running ROM_EXT will cause ROM_EXT boot slot to not change after update
- Check that the application security version is allowed by current minimum BL0 security version, and fail early if it isn't
- Do not fail DFU check if the running ROM_EXT version is not the same as the one in the image being checked in cases where this is expected

### DIFF
--- a/examples/htool_dfu.c
+++ b/examples/htool_dfu.c
@@ -47,8 +47,9 @@ static int dfu_update_count(struct opentitan_image_version* desired_romext,
   struct opentitan_image_version* staged_app = &resp->app.slots[app_stage_slot];
 
   bool booted_needs_update =
-      !libhoth_ot_version_eq(booted_app, desired_app) ||
-      !libhoth_ot_version_eq(booted_romext, desired_romext);
+      (libhoth_ot_app_version_cmp_for_update(booted_app, desired_app) != 0) ||
+      (libhoth_ot_rom_ext_version_cmp_for_update(booted_romext,
+                                                 desired_romext) < 0);
 
   if (booted_needs_update) {
     printf(
@@ -58,8 +59,9 @@ static int dfu_update_count(struct opentitan_image_version* desired_romext,
   }
 
   bool staging_needs_update =
-      !libhoth_ot_version_eq(staged_app, desired_app) ||
-      !libhoth_ot_version_eq(staged_romext, desired_romext);
+      (libhoth_ot_app_version_cmp_for_update(staged_app, desired_app) != 0) ||
+      (libhoth_ot_rom_ext_version_cmp_for_update(staged_romext,
+                                                 desired_romext) < 0);
 
   if (staging_needs_update) {
     printf(
@@ -152,6 +154,16 @@ int htool_dfu_update(const struct htool_invocation* inv) {
     goto cleanup2;
   }
 
+  if (!force && (desired_app.security_version < resp.bl0_min_sec_ver)) {
+    fprintf(
+        stderr,
+        "Desired application firmware security version %u is less than "
+        "currently enforced minimum application firmware security version %u\n",
+        desired_app.security_version, resp.bl0_min_sec_ver);
+    retval = -1;
+    goto cleanup2;
+  }
+
   int update_cnt =
       force ? 2 : dfu_update_count(&desired_rom_ext, &desired_app, &resp);
 
@@ -173,7 +185,8 @@ int htool_dfu_update(const struct htool_invocation* inv) {
       goto cleanup2;
     }
 
-    if (!libhoth_ot_boot_slot_eq(&resp, &desired_rom_ext, &desired_app)) {
+    if (!libhoth_ot_check_update_successful(&resp, &desired_rom_ext,
+                                            &desired_app)) {
       fprintf(stderr, "Boot slot is wrong after dfu update %d\n", i);
       libhoth_print_dfu_error(dev, &resp, LIBHOTH_OK);
       retval = -1;
--- a/protocol/opentitan_version.c
+++ b/protocol/opentitan_version.c
@@ -76,17 +76,20 @@ int libhoth_extract_ot_bundle(const uint8_t* image, size_t image_size,
   static_assert(
       OPENTITAN_OFFSET_VERSION_MINOR >= OPENTITAN_OFFSET_VERSION_MAJOR,
       "Version minor offset must be >= major offset");
+  static_assert(
+      OPENTITAN_OFFSET_VERSION_SECURITY >= OPENTITAN_OFFSET_VERSION_MINOR,
+      "Security version offset must be >= minor version offset");
 
   // Checks that the image offset doesn't cause
   // the offset calculations to read beyond the image buffer.
-  if ((offset + OPENTITAN_OFFSET_APP_FW + OPENTITAN_OFFSET_VERSION_MINOR + 4) >
-      image_size) {
+  if ((offset + OPENTITAN_OFFSET_APP_FW + OPENTITAN_OFFSET_VERSION_SECURITY +
+       4) > image_size) {
     fprintf(stderr, "Image offset %u is invalid", offset);
     return -1;
   }
   // Checks that the image offset won't overflow the offset calculations
-  if ((offset + OPENTITAN_OFFSET_APP_FW + OPENTITAN_OFFSET_VERSION_MINOR + 4) <
-      offset) {
+  if ((offset + OPENTITAN_OFFSET_APP_FW + OPENTITAN_OFFSET_VERSION_SECURITY +
+       4) < offset) {
     fprintf(stderr, "Image offset %u caused integer overflow", offset);
     return -1;
   }
@@ -99,6 +102,11 @@ int libhoth_extract_ot_bundle(const uint8_t* image, size_t image_size,
                    image[offset + OPENTITAN_OFFSET_VERSION_MINOR + 1] << 8 |
                    image[offset + OPENTITAN_OFFSET_VERSION_MINOR + 2] << 16 |
                    image[offset + OPENTITAN_OFFSET_VERSION_MINOR + 3] << 24;
+  rom_ext->security_version =
+      image[offset + OPENTITAN_OFFSET_VERSION_SECURITY] |
+      image[offset + OPENTITAN_OFFSET_VERSION_SECURITY + 1] << 8 |
+      image[offset + OPENTITAN_OFFSET_VERSION_SECURITY + 2] << 16 |
+      image[offset + OPENTITAN_OFFSET_VERSION_SECURITY + 3] << 24;
 
   // Extract the offset that contains the APP version information
   // We will have the desired APP version be stored on slot index 0 and keep
@@ -112,6 +120,11 @@ int libhoth_extract_ot_bundle(const uint8_t* image, size_t image_size,
                image[offset_app + OPENTITAN_OFFSET_VERSION_MINOR + 1] << 8 |
                image[offset_app + OPENTITAN_OFFSET_VERSION_MINOR + 2] << 16 |
                image[offset_app + OPENTITAN_OFFSET_VERSION_MINOR + 3] << 24;
+  app->security_version =
+      image[offset_app + OPENTITAN_OFFSET_VERSION_SECURITY] |
+      image[offset_app + OPENTITAN_OFFSET_VERSION_SECURITY + 1] << 8 |
+      image[offset_app + OPENTITAN_OFFSET_VERSION_SECURITY + 2] << 16 |
+      image[offset_app + OPENTITAN_OFFSET_VERSION_SECURITY + 3] << 24;
 
   return 0;
 }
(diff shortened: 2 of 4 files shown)


protocol/opentitan_version.c
```diff
@@ -231,7 +236,66 @@ bool libhoth_ot_staged_slot_eq(
...
+  if ((desired_rom_ext_version == NULL) || (current_versions_resp == NULL)) {
+    return true;
+  }
+  const struct opentitan_image_version* current_rom_ext_version =
+      libhoth_ot_boot_romext(current_versions_resp);
+  // For booting ROM_EXT, ROM will give priority to slot with greater security
+  // version first, and then greater major version, and then greater minor
+  // version
+  // ([ref](https://github.com/lowRISC/opentitan/blob/acb2179c7f4b59d4ee2118fb24cfb4f97d920dc9/sw/device/silicon_creator/rom/boot_policy.c#L18-L41))
+  if (current_rom_ext_version->security_version >
```
korran: This might be easier to understand with a `int opentitan_version_cmp(const opentitan_image_version* a, const opentitan_image_version* b)` function, which can be easily unit tested.

korran: I guess that would be libhoth_ot_version_cmp(), which libhoth_ot_version_eq() can delegate to...

korran: interestingly, libhoth_ot_version_eq() doesn't currently look at security version...

protocol/opentitan_version.c
```diff
@@ -214,9 +214,14 @@ const struct opentitan_image_version* libhoth_ot_staged_romext(
...
     const struct opentitan_image_version* desired_romext,
-    const struct opentitan_image_version* desired_app) {
-  return libhoth_ot_version_eq(libhoth_ot_boot_romext(resp), desired_romext) &&
-         libhoth_ot_version_eq(libhoth_ot_boot_app(resp), desired_app);
+    const struct opentitan_image_version* desired_app,
+    const bool expect_desired_rom_ext) {
+  bool result = libhoth_ot_version_eq(libhoth_ot_boot_app(resp), desired_app);
+  if (expect_desired_rom_ext) {
+    result &=
+        libhoth_ot_version_eq(libhoth_ot_boot_romext(resp), desired_romext);
```
stevenportley: Are you sure we want to complete omit check the ROM_EXT version in this case?  Would it make more sense to instead check that the ROM_EXT version hasn't changed?

korran: How could it change? the version that booted previously is pinned in flash and the ROM will refuse to boot the old one we just flashed...

xorptr: > Would it make more sense to instead check that the ROM_EXT version hasn't changed

This function is used both at end of DFU update procedure, and during post-update procedure to check whether the running firmware matches what is expected (from the file). I can check that ROM_EXT version hasn't changes at end of the update procedure, but not for the other scenario (since I cannot query the version before the update in that case)

protocol/opentitan_version.c
xorptr: I don't think the security version check here matters for the application firmware. The [boot policy check for application firmware](https://github.com/lowRISC/opentitan/blob/567855c952c833f71ca707c12945d76bd848d600/sw/device/silicon_creator/rom_ext/rom_ext_boot_policy.c#L16-L29) seems to only consult `primary_bl0_slot` (called [here](https://github.com/lowRISC/opentitan/blob/567855c952c833f71ca707c12945d76bd848d600/sw/device/silicon_creator/rom_ext/rom_ext.c#L389-L390)). And then [`rom_ext_verify` call](https://github.com/lowRISC/opentitan/blob/567855c952c833f71ca707c12945d76bd848d600/sw/device/silicon_creator/rom_ext/rom_ext.c#L399) just checks against `min_security_version_bl0` [here](https://github.com/lowRISC/opentitan/blob/567855c952c833f71ca707c12945d76bd848d600/sw/device/silicon_creator/rom_ext/rom_ext_boot_policy.c#L75-L80) (called from [here](https://github.com/lowRISC/opentitan/blob/567855c952c833f71ca707c12945d76bd848d600/sw/device/silicon_creator/rom_ext/rom_ext_verify.c#L28-L29)).

Same concerns apply for application major and minor firmware. My understanding is that which application firmware version to boot depends primarily on `primary_bl0_slot`, so using this logic in checking application firmware version would be incorrect.

That's why I limited the logic to ROM_EXT only (`expect_active_rom_ext_change_with_update`). Even with ROM_EXT, this only helps with comparison against the active ROM_EXT (currently running). The staging ROM_EXT can have a different major and minor version. For eg:
- Before update: Active ROM_EXT is 0.111 with security version 2. Staging ROM_EXT is 0.111 with security version 2
- Update to image with ROM_EXT 0.110 with security version 0
- After update: Active ROM_EXT is 0.111 with security version 2. Staging ROM_EXT is 0.110 with security version 0

I am not sure if putting this logic in a generic `opentitan_version_cmp` function would be helpful.

protocol/opentitan_version_test.cc
```diff
@@ -142,43 +142,173 @@ TEST_F(LibHothTest, opentitan_image_compare_test) {
...
 
+TEST_F(LibHothTest, TestLibhothUpdateComplete) {
   struct opentitan_image_version romext = {
       .major = 6,
       .minor = 111,
       .security_version = 7,
       .timestamp = 1234,
       .measurement = {},
   };
+  struct opentitan_image_version romext_security_version_larger = {
```
korran, line 259: This example also increases the minor version, so how would we know that it is the security version that is being checked?

protocol/opentitan_version.c
korran: My point is that this code is larger and harder to review correctness (and test) than it would be if it was split up. For example, why do we need to have have 3 separate printf's that mostly say the same thing?

```
int opentitan_rom_ext_version_cmp(const opentitan_image_version* a, const opentitan_image_version* b) {
    if (a->security_version != b->security_version) {
        return a->security_version > b->security_version ? 1 : -1;
    }
    if (a->major != b->major) {
        return a->major > b->major ? 1 : -1;
    }
    if (a->minor != b->minor) {
        return a->minor > b->minor ? 1 : -1;
    }
    return 0;
}

// somewhere else

if (opentitan_version_cmp(current_rom_ext_version, desired_rom_ext_version) >= 0) {
  fprintf("ROM_EXT already running version (%d.%d.%d) later than or equal to desired (%d.%d.%d)", ...);
  return false;
}

```

Also, the comments seem way more verbose than necessary and don't seem to be related to the decision of "does the ROM_EXT need to be updated?".

protocol/opentitan_version.c
```diff
@@ -231,7 +236,69 @@ bool libhoth_ot_staged_slot_eq(
...
+  // Failure during this check will cause ROM to try the other ROM_EXT with the
+  // same check applied to it
+  // [here](https://github.com/lowRISC/opentitan/blob/acb2179c7f4b59d4ee2118fb24cfb4f97d920dc9/sw/device/silicon_creator/rom/rom.c#L761-L785).
+  // So it is expected that the running ROM_EXT has security version greater
+  // than or equal to the minimum ROM_EXT security version. In such a case, the
+  // conditions above will already handle firmware update to image with ROM_EXT
+  // version less than the minimum ROM_EXT security version (in the ROM_EXT
+  // security version check between desired and currently running ROM_EXT)
+
+  return true;
```
korran: If desired_rom_ext_version == current_versions_resp, shouldn't this function return false, indicating that the version won't change?

protocol/opentitan_version.c
```diff
@@ -231,7 +236,69 @@ bool libhoth_ot_staged_slot_eq(
...
+    const bool expect_desired_rom_ext) {
+  // In case of ROM_EXT update not occurring due to valid reasons (see
+  // `expect_active_rom_ext_change_with_update`), the staging side will still be
+  // running the ROM_EXT from the update package
+  return libhoth_ot_boot_slot_eq(resp, desired_romext, desired_app,
+                                 expect_desired_rom_ext) &&
          libhoth_ot_staged_slot_eq(resp, desired_romext, desired_app);
 }
+
+bool expect_active_rom_ext_change_with_update(
```
korran: This function name isn't very clear. `rom_ext_needs_update()` might be clearer?

protocol/opentitan_version.c
```diff
@@ -231,7 +236,69 @@ bool libhoth_ot_staged_slot_eq(
...
+    const struct opentitan_image_version* desired_rom_ext_version,
+    const struct opentitan_get_version_resp* current_versions_resp) {
+  if ((desired_rom_ext_version == NULL) || (current_versions_resp == NULL)) {
+    return true;
```
korran: Wouldn't it be better to crash if this ever happens? I don't think there's an obvious "true" or "false" return value in this scenario.

protocol/opentitan_version.h
```diff
@@ -135,15 +135,21 @@ const struct opentitan_image_version* libhoth_ot_staged_romext(
 bool libhoth_ot_boot_slot_eq(
     const struct opentitan_get_version_resp* resp,
     const struct opentitan_image_version* desired_romext,
-    const struct opentitan_image_version* desired_app);
+    const struct opentitan_image_version* desired_app,
+    const bool expect_desired_rom_ext);
```
korran: Rather than plumbing an awkward bool parameter through all these layers, why don't we rename this function `libhoth_ot_was_update_successful`, and have it check that installed rom-ext is the same or newer than the desired rom-ext?

protocol/opentitan_version_test.cc
```diff
@@ -142,43 +142,173 @@ TEST_F(LibHothTest, opentitan_image_compare_test) {
   resp.app.booted_slot = kOpentitanBootSlotB;
```
korran, line 185: None of these test cases seem to call `expect_active_rom_ext_change_with_update()`: is this what you intended?

protocol/opentitan_version.c
xorptr: Renaming to match suggestion in https://github.com/google/libhoth/pull/253/changes#r3306650233

protocol/opentitan_version.c
xorptr: Ack.

protocol/opentitan_version.c
xorptr: Changing the function to match the suggestion in https://github.com/google/libhoth/pull/253/changes#r3306650233

This function currently checks that the active rom_ext version is expected to change after the update or not. With `desired_rom_ext_version == current_versions_resp`, the version would not change.

protocol/opentitan_version.h
xorptr: Ack. That's a good suggestion. I was trying to re-use the existing functions without much changes. But I will change the functions to make the logic more readable.

protocol/opentitan_version_test.cc
xorptr: Yes. I missed that. Thanks for catching that.

protocol/opentitan_version_test.cc
xorptr: good catch! I mean to lower the major or minor version. Thanks

protocol/opentitan_version.c
xorptr: Ack. Changed the implementation to what is suggested in https://github.com/google/libhoth/pull/253/changes#r3306650233

xorptr:
~dfu check fails because current logic does not extract security version from firmware update package~ Fixed now

examples/htool_dfu.c
```diff
@@ -47,8 +47,9 @@ static int dfu_update_count(struct opentitan_image_version* desired_romext,
   struct opentitan_image_version* staged_app = &resp->app.slots[app_stage_slot];
 
   bool booted_needs_update =
-      !libhoth_ot_version_eq(booted_app, desired_app) ||
-      !libhoth_ot_version_eq(booted_romext, desired_romext);
+      (libhoth_ot_app_version_cmp_for_update(booted_app, desired_app) != 0) ||
+      (libhoth_ot_rom_ext_version_cmp_for_update(booted_romext,
+                                                 desired_romext) == -1);
```
korran: nit: idiomatically use of *_cmp() functions is `< 0` rather than `== -1`

When reading the code you can think of `< 0` to mean `booted_romext < desired_romext`.

protocol/opentitan_version.c
```diff
@@ -211,27 +251,34 @@ const struct opentitan_image_version* libhoth_ot_staged_romext(
   return &resp->rom_ext.slots[rom_ext_stage_slot];
 }
 
-bool libhoth_ot_boot_slot_eq(
+bool libhoth_ot_check_boot_slot_after_update(
```
korran: nit: Can this name make it clear what retval=false and retval=true mean?
For example, with `libhoth_ot_boot_slot_update_successful()` it is clear that retval=true means the update succeeded.

protocol/opentitan_version.c
```diff
@@ -211,27 +251,34 @@ const struct opentitan_image_version* libhoth_ot_staged_romext(
...
     const struct opentitan_get_version_resp* resp,
     const struct opentitan_image_version* desired_romext,
     const struct opentitan_image_version* desired_app) {
-  return libhoth_ot_version_eq(libhoth_ot_boot_romext(resp), desired_romext) &&
-         libhoth_ot_version_eq(libhoth_ot_boot_app(resp), desired_app);
+  return (libhoth_ot_app_version_cmp_for_update(libhoth_ot_boot_app(resp),
+                                                desired_app) == 0) &&
+         (libhoth_ot_rom_ext_version_cmp_for_update(
+              libhoth_ot_boot_romext(resp), desired_romext) != -1);
```
korran: nit: `>= 0`?

korran (approved):

protocol/opentitan_version.c
xorptr: Done

protocol/opentitan_version.c
xorptr: Renamed to `libhoth_ot_check_update_successful` (since putting `boot_slot` in the function name seems to add implementation details)

examples/htool_dfu.c
xorptr: Done

PR #254, opened 2026-05-29
[protocol] Convert hello command and libhoth to use the new error

esnguyen:
As a way to centeralize the way we report error through different domains, we are using hello command as a poc. This should make a temporary libhoth exec to be non instrusive and allow for changes.


protocol/status.h
```diff
@@ -57,28 +66,28 @@ enum hoth_host_space {
 // Hoth error code: Low 16 bits of the 32-bit Base Error Code
 // Firmware Error
 enum hoth_fw_error_status {
```
stevenportley, line 70: I assume this is intended to replace the hoth_status in host_cmd.h?

One thing we should be careful of is that host commands can return one of two different types of errors:
- Dauntless only supports the subset you have here, a u16
- Earlgrey also supports a different type of error that u32 and provides a lot more context


We might want to reserve a 16bit prefix for these old Dauntless error codes, and add it onto the error code whenever we don't get the full u32.  We can probably just change HTOOL_ERROR_HOST_COMMAND_START to do this since I don't see it used anywhere outside of libhoth

protocol/host_cmd.c
```diff
@@ -162,10 +162,43 @@ static int validate_ec_response_header(
...
+    return HTOOL_ERROR_HOST_COMMAND_START + code;
+  }
+
+  return -1;
+}
+
+// libhoth_hostcmd_exec_v2 is the new core implementation that returns a
+// rich 64-bit libhoth_error. It should be used for all new commands and
+// when migrating existing commands to the centralized status reporting system.
+libhoth_error libhoth_hostcmd_exec_v2(struct libhoth_device* dev,
```
stevenportley, line 200: I assume the plan is to complete the migrate to this function and then to just rename everything back to `libhoth_hostcm_exec`?  I don't think we have any consumers of this API outside of  this repo so that should be safe.

stevenportley (approved):

protocol/host_cmd.c
esnguyen: @stevenportley Yes this is exactly the case. this will allow for support for the new function without breaking anybody else

PR #255, opened 2026-06-01
[protocol] Add the set mauv record support in htool

esnguyen:
As the haven/dauntless firmware has support to set the mauv record in memory, we are adding the support in the command to allow htool to sends a record.

### DIFF
--- a/examples/BUILD
+++ b/examples/BUILD
@@ -46,6 +46,9 @@ cc_library(
 cc_library(
     name = "htool_header",
     hdrs = ["htool.h"],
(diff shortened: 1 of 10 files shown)


ckungler (approved):
Approving the code itself. I'm not very familiar with MAUV process

esnguyen:
> Approving the code itself. I'm not very familiar with MAUV process

Ack! This is a mirror change so that is in haventool. Just construction of the packet to be sent with unit test. minor impact

PR #258, opened 2026-06-03
Add top level libhoth bazel target

stevenportley:
This will be used to consolidate all libhoth features for easy consumption from other projects. It also exports libhoth transport header files to both libhoth/transport/* and transport/*.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -20,6 +20,37 @@ genrule(
     tools = [":print_version_header.sh"],
 )
 
+cc_library(
+    name = "libhoth_transports_headers",
+    hdrs = [
+      "//transports:headers",
+    ],
+    include_prefix = "libhoth/transports",
+    strip_include_prefix = "transports",
+)
+
+cc_library(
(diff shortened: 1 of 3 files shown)


BUILD
```diff
@@ -20,6 +20,37 @@ genrule(
...
+    name = "libhoth_transports_headers_legacy",
+    hdrs = [
+      "//transports:headers",
+    ],
+    include_prefix = "transports",
+    strip_include_prefix = "transports",
+)
+
+cc_library(
+    name = "libhoth",
```
xorptr, line 42: nit: should this be `libhoth_transports`? libhoth can also provide protocol layer, so I think using a more specific name here might be better.

xorptr (approved):

BUILD
stevenportley: I was planning on adding all of protocol/ to this library eventually, and using just this target from both htool and hothd.  We probably want to wait for the error code rework to go through before we do that though.

PR #261, opened 2026-06-08
[protocol] Convert dfu_check to use the new compact error format

esnguyen:
Migrates libhoth_dfu_check off the legacy integer return type onto libhoth_error and libhoth_hostcmd_exec_v2.


protocol/dfu_check_test.cc
```diff
@@ -93,7 +93,15 @@ TEST_F(LibHothTest, dfu_check_fail) {
   mock_response.app.slots[1].major = 0;
   mock_response.app.slots[1].minor = 20;
 
-  EXPECT_EQ(
+  EXPECT_NE(
```
stevenportley: Maybe we should EXPECT_EQ with the expected error code now that we are getting meaningful error codes back?

stevenportley: Might want to also add a few more tests for other possible failure scenarios, if convenient.

protocol/dfu_check.c
```diff
@@ -99,9 +107,10 @@ int libhoth_dfu_check(struct libhoth_device* const dev, const uint8_t* image,
     fprintf(
         stderr,
         "Error: Mismatch detected between the current and desired versions.\n");
-    libhoth_print_dfu_error(dev, resp, LIBHOTH_OK);
-    return -1;
+    libhoth_print_dfu_error(dev, resp, HOTH_SUCCESS);
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
```
stevenportley: We should construct something meaningful here.  Maybe allocate a (or a few) unique error codes for libhoth dfu check failures?  If we ever hit this, it means that the dfu update passed, but the dfu check failed, which is interesting information we need to give back to the caller for debug.

We would probably want to know where the mismatch was, either between ROM_EXT checks or App version checks, so we probably want to return an appropriate error code inside `libhoth_update_complete` and just propagate that back from here.

protocol/dfu_check.c
stevenportley, line 93: Maybe also change libhoth_extract_ot_bundle to return libhoth_error and then propagate that here?

stevenportley (approved):

protocol/dfu_check.c
esnguyen: @stevenportley Will follow up on this with another bug recommended offline.

PR #262, opened 2026-06-09
[protocol] Convert dfu_hostcmd to use the new compact error format

esnguyen:
Migrates libhoth_dfu_update off the legacy integer return type onto libhoth_error and libhoth_hostcmd_exec_v2.

### DIFF
--- a/protocol/dfu_hostcmd.c
+++ b/protocol/dfu_hostcmd.c
@@ -24,14 +24,21 @@ static int generate_random_nonce(struct hoth_dfu_session_id* session_id) {
   return 0;
 }
 
-int libhoth_dfu_update(struct libhoth_device* dev, const uint8_t* image,
-                       size_t image_size, uint32_t complete_flags) {
+libhoth_error libhoth_dfu_update(struct libhoth_device* dev,
+                                 const uint8_t* image, size_t image_size,
+                                 uint32_t complete_flags) {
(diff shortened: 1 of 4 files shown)


protocol/dfu_hostcmd.c
```diff
@@ -72,17 +80,22 @@ int libhoth_dfu_update(struct libhoth_device* dev, const uint8_t* image,
...
+            "DFU complete failed with error code: 0x%016lx; ignoring as the "
             "chip may have already restarted.\n",
-            ret);
+            err);
   }
 
   // TODO: Wait for chip to come back and confirm version
   usleep(LIBHOTH_REBOOT_DELAY_MS * 1000);
-  return libhoth_device_reconnect(dev);
+  int ret = libhoth_device_reconnect(dev);
```
stevenportley, line 95: We should also plan to migrate all of the transport functions to return libhoth_error too, but they are actually used in hothd and maybe some other places so the migration might be a little more challenging.

stevenportley (approved):

protocol/dfu_hostcmd.c
esnguyen: @stevenportley yes this is true. currently working on finding a way to make sure the these changes to transport can be rolled out in a phase with little to no impact to extern users.

PR #267, opened 2026-06-09
[protocol] Convert payload_status to use the new compact error format

esnguyen:
Migrates libhoth_payload_status off the legacy integer return type onto libhoth_error and libhoth_hostcmd_exec_v2.

### DIFF
--- a/protocol/payload_status.c
+++ b/protocol/payload_status.c
@@ -18,15 +18,20 @@
 
 #include "host_cmd.h"
 
-int libhoth_payload_status(struct libhoth_device* dev,
-                           struct payload_status* payload_status) {
+libhoth_error libhoth_payload_status(struct libhoth_device* dev,
+                                     struct payload_status* payload_status) {
+  if (payload_status == NULL) {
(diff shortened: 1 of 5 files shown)


protocol/payload_status.c
```diff
@@ -18,26 +18,32 @@
 
 #include "host_cmd.h"
 
-int libhoth_payload_status(struct libhoth_device* dev,
-                           struct payload_status* payload_status) {
+libhoth_error libhoth_payload_status(struct libhoth_device* dev,
+                                     struct payload_status* payload_status) {
+  if (payload_status == NULL) {
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
```
stevenportley, line 24: nit: You use this definition in a bunch of different places.  Probably worthwhile to define it directly in status.h or something and just return it.

stevenportley (approved):

protocol/payload_status.c
esnguyen: @stevenportley I like this approach but I think I would need to do this on another PR that focuses on leveraging a defines from commonly used error in respects to their context and space

PR #268, opened 2026-06-09
[protocol] Convert reboot to use the new compact error format

esnguyen:
Migrates libhoth_reboot off the legacy integer return type onto libhoth_error and libhoth_hostcmd_exec_v2.

### DIFF
--- a/protocol/reboot.c
+++ b/protocol/reboot.c
@@ -14,12 +14,15 @@
 
 #include "reboot.h"
 
+#include <stddef.h>
+
 #include "host_cmd.h"
 
-int libhoth_reboot(struct libhoth_device* dev, enum hoth_reboot_cmd cmd) {
+libhoth_error libhoth_reboot(struct libhoth_device* dev,
+                             enum hoth_reboot_cmd cmd) {
   struct hoth_params_reboot_ec req = {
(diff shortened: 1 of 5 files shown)


protocol/reboot.c
```diff
@@ -14,12 +14,19 @@
...
 
+#include <stddef.h>
+
 #include "host_cmd.h"
 
-int libhoth_reboot(struct libhoth_device* dev, enum hoth_reboot_cmd cmd) {
+libhoth_error libhoth_reboot(struct libhoth_device* dev,
+                             enum hoth_reboot_cmd cmd) {
+  if (cmd != HOTH_REBOOT_COLD && cmd != HOTH_REBOOT_WARM) {
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
```
stevenportley: Is this really necessary? It's already of type `enum hoth_reboot_cmd`

stevenportley (approved):

protocol/reboot.c
esnguyen: Not really necessary. I can have this guard removed

PR #272, opened 2026-06-09
[protocol] Convert authz_record to use the new compact error format

esnguyen:
Migrates libhoth_authz_record_erase, libhoth_authz_record_read, libhoth_authz_record_build, and libhoth_authz_record_set off the legacy integer return type onto libhoth_error and
libhoth_hostcmd_exec_v2

### DIFF
--- a/protocol/authz_record_test.cc
+++ b/protocol/authz_record_test.cc
@@ -37,7 +37,7 @@ TEST_F(LibHothTest, authz_erase_test) {
   EXPECT_CALL(mock_, receive)
       .WillOnce(DoAll(CopyResp(&dummy, 0), Return(LIBHOTH_OK)));
 
-  EXPECT_EQ(libhoth_authz_record_erase(&hoth_dev_), LIBHOTH_OK);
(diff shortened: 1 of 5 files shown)


protocol/authz_record_test.cc
```diff
@@ -47,7 +47,7 @@ TEST_F(LibHothTest, authz_erase_test) {
 
   EXPECT_CALL(mock_, receive).WillOnce(DoAll(CopyResp(&dummy, 0), Return(-1)));
 
-  EXPECT_EQ(libhoth_authz_record_erase(&hoth_dev_), -1);
+  EXPECT_NE(libhoth_authz_record_erase(&hoth_dev_), HOTH_SUCCESS);
```
stevenportley: We should explicitly be testing that the correct error is returned, not just that it wasn't successful.

stevenportley (approved):

protocol/authz_record_test.cc
esnguyen: Thats fair. I can add the test to return the correct failure.

PR #275, opened 2026-06-17
Implement V2 transport interface, fallbacks, and migrate hostcmd

esnguyen:


transports/libhoth_device.h
```diff
@@ -84,6 +97,24 @@ int libhoth_claim_device(struct libhoth_device* dev, uint32_t timeout_us);
 
 int libhoth_release_device(struct libhoth_device* dev);
 
+// --- New V2 Helper Functions ---
+libhoth_error libhoth_send_request_v2(struct libhoth_device* dev,
```
stevenportley, line 101: What's the purpose of adding a new set of helper functions here? I presume we're just going to remove them in a few commits once the implementations are migrated?

protocol/host_cmd.c
```diff
@@ -224,24 +224,25 @@ libhoth_error libhoth_hostcmd_exec_v2(struct libhoth_device* dev,
     return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_POSIX,
                                  -status);
   }
-  status = libhoth_send_request(dev, &req, sizeof(req.hdr) + req_payload_size);
-  if (status != LIBHOTH_OK) {
-    fprintf(stderr, "libhoth_send_request() failed: %d\n", status);
-    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
-                                 status);
+  libhoth_error err =
+      libhoth_send_request_v2(dev, &req, sizeof(req.hdr) + req_payload_size);
```
stevenportley, line 228: What's the reason for using the `libhoth_send_request_v2` here? 

I would expect we could just implement the v2 callbacks for each of the transports, then just delete the originals and change the helper interface. 

Maybe I'm not understanding the transition for this?

stevenportley (changes requested):

transports/libhoth_device.h
esnguyen: I think the semantic of v2 "Helper" function is probably confusing the purpose of this function. The new function  libhoth_send_request_v2, ***send_v2, etc. will be retain after some commits. These will be the new function to support the 64 error. In this context, the reason why Gemini labeled it as a helper function is because libhoth_send_request and the other transport API are still the existing and dominate functionally. These v2 are just duplicate to support back comparability for the v1 functions.

protocol/host_cmd.c
esnguyen: @stevenportley This is Gemini response. I think she worded better than I did. lert me know if this makes sense. ```
"That is a very fair point, and in a single-repository project, changing the helper interface directly would indeed be the cleanest approach!
  │
  │ The reason we are introducing  libhoth_send_request_v2  alongside the original is to prevent breaking compilation for external consumers (like
  │ gbmc_havend  in the BMC repo and  hsm_usb_interface  in the Haven HSM repo) which depend on  libhoth  but live in separate repositories.
  │
  │ If we changed the signature of the existing  libhoth_send_request  directly to return  libhoth_error , it would immediately break those external
  │ builds the next time they update their  libhoth  dependency.
  │
  │ By introducing the  _v2  helpers, we allow those external teams to update  libhoth  safely, migrate their code to the V2 APIs at their own pace, and
  │ once everyone is fully migrated, we will submit a cleanup PR to remove the legacy V1 helpers and rename these back to the clean interface."
```

stevenportley: I think this implies that we :
- add the v2 helpers
- change everything using libhoth to use the v2 helpers
- change the original to the new signature
- change everything using libhoth back to original
- delete the v2 helpers.

If this were a heavily used API and the signature change was complicated, I would be tempted to go with this, but the signature change is just int -> uint64_t for the return type.  This is only used in one or two places (from what I can tell), and most uses just compare the return type to LIBHOTH_OK.  

Maybe try just making the signature change and seeing if it compiles okay with hothd before doing all of this?

esnguyen: Awh thanks for clarifying the external user's use-cases. I think this would be an simplified approach that Ill look into.

PR #276, opened 2026-06-23
Refactor transports to use libhoth_error

esnguyen:
Migrates the transport layer in dev/libhoth from legacy 'int' return codes to the new rich 64-bit 'libhoth_error' format. This allows propagating detailed error information (like libusb or POSIX errno) all the way to callers.

Includes:
- Core libhoth_device interface and wrapper updates
- USB, SPI, MTD, and DBUS transport migrations with rich error mapping
- host_cmd.c updates to propagate rich errors directly
- htool_usb and htool_console updates to handle libhoth_error

### DIFF
--- a/examples/htool_console.c
+++ b/examples/htool_console.c
@@ -110,19 +110,20 @@ int htool_console_run(struct libhoth_device* dev,
       // TODO: Read STDIN during this time and buffer it so we can capture
       // quit events even when the console is disconnected. We will also want
       // to tune the reconnect time to match.
-      status = libhoth_device_reconnect(dev);
-      // If USB is down we might fail reconnect, just retry
-      if (status != LIBHOTH_OK) {
+      if (libhoth_device_reconnect(dev) != HOTH_SUCCESS) {
         // Make sure we don't end up in a tight retry loop
         usleep(100 * 1000);
+      } else {
+        status = LIBHOTH_OK;
       }
     }
     // Give an opportunity for other clients to use the interface.
     libhoth_release_device(dev);
     usleep(1000 * opts->yield_ms);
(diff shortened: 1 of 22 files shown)


examples/htool_console.c
```diff
@@ -103,25 +103,29 @@ int htool_console_run(struct libhoth_device* dev,
   // buffer is much smaller than 2GB).
   if (opts->history) offset -= 0x80000000;
   bool quit = false;
+  libhoth_error transport_err = HOTH_SUCCESS;
 
   while (!quit) {
     // Any previous failure should reset all of the USB state before retrying
-    while (status != LIBHOTH_OK) {
+    while (status != LIBHOTH_OK || transport_err != HOTH_SUCCESS) {
```
stevenportley: Do we need to keep both `status` and `transport_err`?

transports/libhoth_usb.c
```diff
@@ -30,16 +30,44 @@
...
+  }
+  return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_USB, HOTH_HOST_SPACE_LIBHOTH,
+                               libhoth_err);
+}
+
+static libhoth_error wrap_usb_err(int ret) {
+  if (ret == 0) {
+    return HOTH_SUCCESS;
+  }
+  if (ret < 0) {
```
stevenportley, line 50: This seems unnecessary?

stevenportley (approved):

transports/libhoth_usb.c
esnguyen: will remove the redundant part

examples/htool_console.c
esnguyen: no we don't. will remove

PR #277, opened 2026-06-23
usb: Add auto-detection to libhoth_usb_get_device

stevenportley:
This allows libhoth_usb_get_device to auto-detect a USB hoth device if NULL is provided for the usb_loc.


esnguyen (approved):
LGTM. simple logic to address auto detect. no major logical issue seen

PR #285, opened 2026-08-03
dfu_hostcmd: Move dfu install to protocol/

stevenportley:
This moves the rest of the DFU update procedure into protocol/ so that it is usable outside of htool.


esnguyen (approved):
LGTM. spoke offline and simple logic change make sense

PR #290, opened 2026-09-04
[protocol] Convert key_rotation to use the new compact error format

esnguyen:
Convert the current legecy int to a new error format to provide a detailed logs of key rotation


protocol/key_rotation.c
```diff
@@ -12,51 +12,61 @@
...
   }
   if (rlen != 0) {
     fprintf(stderr,
             "HOTH_KEY_ROTATION_COMMAND: %d expected exactly %d response "
-            "bytes, got %ld\n",
+            "bytes, got %zu\n",
             command, 0, rlen);
-    return KEY_ROTATION_ERR_INVALID_RESPONSE_SIZE;
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
+                                 LIBHOTH_ERR_FAIL);
```
pbadre: Can we add another HOTH_RES_INVALID_RESPONSE_SIZE and return that here? There is a HOTH_RES_RESPONSE_TOO_BIG but it is not generic enough. The response could be smaller than expected

protocol/key_rotation.c
```diff
@@ -65,30 +75,36 @@ enum key_rotation_err libhoth_key_rotation_get_version(
...
 
   if (rlen != sizeof(*record_version)) {
     fprintf(stderr,
-            "HOTH_KEY_ROTATION_GET_VERSION expected exactly %ld response "
-            "bytes, got %ld\n",
+            "HOTH_KEY_ROTATION_GET_VERSION expected exactly %zu response "
+            "bytes, got %zu\n",
             sizeof(*record_version), rlen);
-    return KEY_ROTATION_ERR_INVALID_RESPONSE_SIZE;
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
```
pbadre: Can we add another HOTH_RES_INVALID_RESPONSE_SIZE and return that here? There is a HOTH_RES_RESPONSE_TOO_BIG but it is not generic enough. The response could be smaller than expected

protocol/key_rotation.c
```diff
@@ -97,29 +113,35 @@ enum key_rotation_err libhoth_key_rotation_get_status(
...
 
   if (rlen != sizeof(*record_status)) {
     fprintf(stderr,
-            "HOTH_KEY_ROTATION_GET_STATUS expected exactly %ld response "
-            "bytes, got %ld\n",
+            "HOTH_KEY_ROTATION_GET_STATUS expected exactly %zu response "
+            "bytes, got %zu\n",
             sizeof(*record_status), rlen);
-    return KEY_ROTATION_ERR_INVALID_RESPONSE_SIZE;
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
```
pbadre: Same as above, HOTH_RES_INVALID_RESPONSE_SIZE

protocol/key_rotation.c
```diff
@@ -128,43 +150,47 @@ enum key_rotation_err libhoth_key_rotation_payload_status(
...
   if (rlen != sizeof(*payload_status)) {
     fprintf(stderr,
-            "HOTH_KEY_ROTATION_PAYLOAD_STATUS expected exactly %ld response "
-            "bytes, got %ld\n",
+            "HOTH_KEY_ROTATION_PAYLOAD_STATUS expected exactly %zu response "
+            "bytes, got %zu\n",
             sizeof(*payload_status), rlen);
-    return KEY_ROTATION_ERR_INVALID_RESPONSE_SIZE;
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
+                                 LIBHOTH_ERR_FAIL);
```
pbadre: Same as above, HOTH_RES_INVALID_RESPONSE_SIZE

protocol/key_rotation.c
```diff
@@ -128,43 +150,47 @@ enum key_rotation_err libhoth_key_rotation_payload_status(
...
   }
   fprintf(stderr, "Initiating key rotation update protocol with libhoth.\n");
-  if (send_key_rotation_request(dev, KEY_ROTATION_RECORD_INITIATE) !=
-      KEY_ROTATION_CMD_SUCCESS) {
+  libhoth_error init_err =
+      send_key_rotation_request(dev, KEY_ROTATION_RECORD_INITIATE);
+  if (init_err != HOTH_SUCCESS) {
     fprintf(stderr, "Failed to initiate key rotation.\n");
-    return KEY_ROTATION_INITIATE_FAIL;
+    return init_err;
```
pbadre: It would make sense to add either KEY_ROTATION_INITIATE_FAIL or generic HTOOL_CMD_INITIATE_FAIL and return that here. Generic one can be reused for payload update as well

protocol/key_rotation.c
```diff
@@ -178,38 +204,45 @@ enum key_rotation_err libhoth_key_rotation_update(struct libhoth_device* dev,
...
   }
   fprintf(stderr, "Finalizing key rotation update.\n");
-  if (send_key_rotation_request(dev, KEY_ROTATION_RECORD_COMMIT) !=
-      KEY_ROTATION_CMD_SUCCESS) {
+  libhoth_error commit_err =
+      send_key_rotation_request(dev, KEY_ROTATION_RECORD_COMMIT);
+  if (commit_err != HOTH_SUCCESS) {
     fprintf(stderr, "Failed to commit key rotation.\n");
-    return KEY_ROTATION_COMMIT_FAIL;
+    return commit_err;
```
pbadre: Same as initiate fail, could add commit fail

protocol/status.h
pbadre, line 95: Could you add an enum `HOTH_RES_INVALID_RESPONSE_SIZE`? Apologies, my earlier comment was not clear. I see `HOTH_RES_RESPONSE_TOO_BIG` which is not generic enough. You can return `HOTH_RES_INVALID_RESPONSE_SIZE` above where I mentioned to return HOTH_RES_INVALID_RESPONSE_SIZE in my comments.

pbadre (approved):

PR #291, opened 2026-09-04
[protocol] Convert jtag to use the new compact error format

esnguyen:
This is the effort to convert the int to the new error format. This case is for the JTAG.

### DIFF
--- a/examples/htool_jtag.c
+++ b/examples/htool_jtag.c
@@ -58,9 +58,11 @@ static int jtag_read_idcode(struct libhoth_device* dev,
   }
 
   uint32_t idcode = 0;
-  int ret = libhoth_jtag_read_idcode(dev, interface_id, clk_idiv, &idcode);
-  if (ret != 0) {
-    return ret;
+  libhoth_error err =
+      libhoth_jtag_read_idcode(dev, interface_id, clk_idiv, &idcode);
+  if (err != HOTH_SUCCESS) {
+    htool_report_error("jtag read_idcode", err);
+    return -1;
   }
 
   printf("IDCODE: 0x%08x\n", idcode);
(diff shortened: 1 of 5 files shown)


protocol/jtag.c
```diff
@@ -31,31 +37,38 @@ int libhoth_jtag_read_idcode(struct libhoth_device* dev, uint8_t interface_id,
...
-    fprintf(stderr,
-            "HOTH_JTAG_OPERATION expected exactly %zu reseponse bytes, got %zu",
-            sizeof(response), response_length);
-    return -1;
+    fprintf(
+        stderr,
+        "HOTH_JTAG_OPERATION expected exactly %zu response bytes, got %zu\n",
+        sizeof(response), response_length);
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
+                                 LIBHOTH_ERR_FAIL);
```
xorptr, line 57: nit: Would it be a good idea to create a different error code for this condition? It can be done in a separate PR if you prefer to do so

protocol/jtag_test.cc
```diff
@@ -69,9 +83,9 @@ TEST_F(LibHothTest, jtag_read_idcode_receive_error) {
   uint32_t received_idcode;
   uint8_t interface_id = 0;
   uint16_t clk_idiv = 0;
-  EXPECT_EQ(libhoth_jtag_read_idcode(&hoth_dev_, interface_id, clk_idiv,
+  EXPECT_NE(libhoth_jtag_read_idcode(&hoth_dev_, interface_id, clk_idiv,
                                      &received_idcode),
-            -1);
+            HOTH_SUCCESS);
```
xorptr, line 88: nit: For the failure cases, do you think checking that a particular error code is returned for the particular scenario in the test would be useful? Or would it be too cumbersome to manage the tests if error codes are tied to particular error scenarios

xorptr (approved):

examples/htool_jtag.c
xorptr: nit: Should this file include `protocol/status.h`?

protocol/jtag.c
esnguyen: @xorptr Yes I think this could be done in a separate PR when I am cleaning through this code. Ill make a note here

protocol/jtag_test.cc
esnguyen: For now. I think having it be placed to granular will be too much to manage. I think if it is common enough failure it should be added.

PR #294, opened 2026-09-08
[protocol] Convert payload_update to use the new compact error format

esnguyen:
Convert the legacy int to the new error format to provide detailed logs for payload_update


protocol/payload_update.c
```diff
@@ -298,73 +311,78 @@ enum payload_update_err libhoth_payload_update_read_chunk(
...
-    if (ret != 0) {
+    int write_ret = libhoth_force_write(fd, buffer, chunk_size);
+    if (write_ret != 0) {
       fprintf(stderr,
               "Failed to write payload during payload read, err code: %d\n",
-              ret);
-      return PAYLOAD_UPDATE_READ_FAIL;
+              write_ret);
+      return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_POSIX,
+                                   errno ? errno : EIO);
```
rkr35: `fprintf` might overwrite `errno`.

Pass `write_ret` directly.

protocol/payload_update.c
```diff
@@ -160,30 +166,32 @@ enum payload_update_err libhoth_payload_update_erase(
...
-                                               uint8_t* image, size_t size,
-                                               bool skip_erase,
-                                               bool binary_file) {
+libhoth_error libhoth_payload_update(struct libhoth_device* dev, uint8_t* image,
+                                     size_t size, bool skip_erase,
+                                     bool binary_file) {
   if (!binary_file && (libhoth_find_image_descriptor(image, size) == NULL)) {
-    return PAYLOAD_UPDATE_BAD_IMG;
+    return LIBHOTH_ERR_CONSTRUCT(HOTH_CTX_CMD_EXEC, HOTH_HOST_SPACE_LIBHOTH,
+                                 LIBHOTH_ERR_INVALID_PARAMETER);
```
rkr35: This seems like a regression. `PAYLOAD_UPDATE_BAD_IMG` was more descriptive than `LIBHOTH_ERR_INVALID_PARAMETER`.

protocol/payload_update.c
```diff
@@ -224,12 +232,13 @@ enum payload_update_err libhoth_payload_update(struct libhoth_device* dev,
...
 
-    int ret = libhoth_hostcmd_exec(
+    libhoth_error ret = libhoth_hostcmd_exec_v2(
         dev, HOTH_CMD_BOARD_SPECIFIC_BASE + HOTH_PRV_CMD_HOTH_PAYLOAD_UPDATE, 0,
         buffer, sizeof(request) + chunk_size, NULL, 0, NULL);
-    if (ret != 0) {
-      fprintf(stderr, "Error code from hoth: %d\n", ret);
-      return PAYLOAD_UPDATE_FLASH_FAIL;
+    if (ret != HOTH_SUCCESS) {
+      fprintf(stderr, "Error code from hoth: 0x%016llx\n",
```
rkr35: Let's use `PRIx64` here and elsewhere.

protocol/payload_update_test.cc
```diff
@@ -159,10 +160,10 @@ TEST_F(LibHothTest, payload_update_erase_fail) {
   uint8_t buffer[4096] = {0};
   std::memcpy(buffer, &kMagic, sizeof(kMagic));
 
-  EXPECT_EQ(libhoth_payload_update(&hoth_dev_, buffer, sizeof(buffer),
+  EXPECT_NE(libhoth_payload_update(&hoth_dev_, buffer, sizeof(buffer),
                                    /*skip_erase=*/false,
                                    /*binary_file=*/false),
-            PAYLOAD_UPDATE_ERASE_FAIL);
+            HOTH_SUCCESS);
```
rkr35: These assertion changes are regressions.

They don't check the specific reason why a payload update is failing, e.g., `PAYLOAD_UPDATE_ERASE_FAIL` vs `PAYLOAD_UPDATE_BAD_IMG`.

_Any_ payload update failure will pass all of these tests.

rkr35 (changes requested):

protocol/payload_update.c
esnguyen: Awh! good catch

protocol/payload_update_test.cc
esnguyen: Yeah this one was an interesting find. before we didn't have the real EC error propagate back into Libhoth so we have these error as certain point. Now that we do, it subtle changes how these test works

examples/htool_payload_update.c
```diff
@@ -73,30 +74,14 @@ int htool_payload_update(const struct htool_invocation* inv) {
...
-    case PAYLOAD_UPDATE_FINALIZE_FAIL:
-      fprintf(stderr, "Error when finalizing.\n");
-      break;
-    case PAYLOAD_UPDATE_IMAGE_NOT_SECTOR_ALIGNED:
-      fprintf(stderr, "Payload image is not sector-aligned.\n");
-      break;
-    default:
-      break;
+  if (payload_update_status != HOTH_SUCCESS) {
+    htool_report_error("payload_update", payload_update_status);
```
rkr35, line 93: To be clear, what error message does the user see with this change?

For example, for `PAYLOAD_UPDATE_BAD_IMG`, `htool` currently prints `"Not a valid Titan image"`. What does this change print in this case?

esnguyen: `Error: 'payload_update' failed (0x001E000500000010): [CMD_EXEC][LIBHOTH][0x00000010]` Basically is doesn't have the string encoding. It has it for EC but I'll add it for Libhoth conversion as well

rkr35 (approved):
Try out the payload commands with these changes and make sure the UX is helpful and not verbose, e.g., the same error is not printed three times.

PR #295, opened 2026-09-14
[examples] add htool interface for mauv effective

esnguyen:
Since the firware support to get the effective mauv version. We just need to conenct the htool interface to return back the data collected.


examples/htool_mauv.c
```diff
@@ -63,7 +63,20 @@ int htool_mauv_compiled(const struct htool_invocation* inv) {
 }
 
 int htool_mauv_effective(const struct htool_invocation* inv) {
-  // TODO: support FW MAUV effective once it's implemented in firmware.
+  struct libhoth_device* dev = htool_libhoth_device();
+  if (!dev) {
+    return -1;
+  }
+
+  struct hoth_response_mauv mauv;
```
mfield4: Should probably zero init this struct.

mfield4 (approved):

PR #296, opened 2026-09-15
[htool] Add missing RoT provisioning commands

willyzha:
Add provisioning subcommands to htool:
- provisioning write: Write a provisioning blob to RoT storage from a binary file.
- provisioning store_secrets: Store encrypted secrets to the RoT from a binary file.
- provisioning load_mldsa_key: Load an ML-DSA key onto the RoT from a binary file.
- provisioning validate_and_sign: Add optional --output flag to save the signed payload.


examples/host_commands.h
```diff
@@ -79,9 +79,12 @@ struct hoth_response_flash_spi_info {
  */
 #define HOTH_BASE_CMD(cmd) (HOTH_CMD_BOARD_SPECIFIC_BASE + (cmd))
 
-/* The major command identifier for the Provisioning Log read host command. */
+/* The major command identifier for the Provisioning Log host command. */
```
xorptr: nit: I think the host command code macros are being moved to respective protocol files ([example 1](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/protocol/chipinfo.h#L27), [example 2](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/protocol/payload_status.h#L27), [example 3](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/protocol/opentitan_version.h#L31)). Would it be a good idea to do so for provisioning commands as well?

examples/htool.c
```diff
@@ -2059,6 +2059,36 @@ static const struct htool_cmd CMDS[] = {
                  .desc = "The signed cert file."},
                 {}},
     },
+    {
+        .verbs = (const char*[]){"provisioning", "store_secrets", NULL},
+        .desc = "Store secrets encrypted to the provisioning encryption key",
```
xorptr: nit: Store secrets encrypted *with* the provisioning encryption key?

examples/htool.c
```diff
@@ -2059,6 +2059,36 @@ static const struct htool_cmd CMDS[] = {
...
+                {}},
+    },
+    {
+        .verbs = (const char*[]){"provisioning", "load_mldsa_key", NULL},
+        .desc = "Load the ML-DSA-44 public key to the RoT",
+        .func = htool_provisioning_load_mldsa_key,
+        .params =
+            (const struct htool_param[]){
+                {HTOOL_FLAG_VALUE, .name = "key", .default_value = "",
+                 .desc = "Path to the 1312-byte ML-DSA-44 public key file."},
```
xorptr: nit: Would it be a good idea to clarify whether this needs to be in PEM or  DER format, or just raw binary file with 1312 bytes?

examples/htool_provisioning.c
```diff
@@ -216,11 +232,13 @@ int htool_validate_and_sign(const struct htool_invocation* inv) {
...
 
-  output_ptr = fopen(output_file, "wb");
-  if (output_ptr == NULL) {
-    printf("Error: %s, when attempting to open file: %s\n", strerror(errno),
-           output_file);
-    goto cleanup;
+  if (strlen(output_file) > 0) {
+    output_ptr = fopen(output_file, "wb");
+    if (output_ptr == NULL) {
+      printf("Error: %s, when attempting to open file: %s\n", strerror(errno),
```
xorptr: nit: I think that rest of the code prints error messages to `stderr` ([example 1](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/protocol/authz_record.c#L67-L68), [example 2](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/protocol/dfu_check.c#L97-L99), [example 3](https://github.com/google/libhoth/blob/e04b97e5d4f831ef73d0bb5efbc4ccc0d5e5e15a/examples/htool_payload_update.c#L66)).

It is probably okay to leave this as it is for now to stay consistent with rest of the file, and change it later

examples/htool_provisioning.h
```diff
@@ -18,14 +18,62 @@
...
+
+enum hoth_key_provisioning_command {
+  HOTH_KEY_PROVISIONING_GET_ENCRYPTION_KEY = 0,
+  HOTH_KEY_PROVISIONING_STORE_SECRETS = 1,
+  HOTH_KEY_PROVISIONING_LOAD_MLDSA_PUBLIC_KEY = 2,
+};
+
+struct hoth_key_provisioning_request_header {
+  uint8_t version;
+  uint8_t command;
```
xorptr: nit: Would it be a good idea to add a comment that this should be one of the variants in `enum hoth_key_provisioning_command`?

examples/htool.c
willyzha: done

examples/htool_provisioning.c
willyzha: I updated this file to print errors to stderr.

examples/host_commands.h
willyzha: Done.

examples/htool.c
willyzha: should be raw 1312 bytes.

willyzha: updated

examples/htool_provisioning.h
willyzha: done.

